PR #14, opened 2019-08-16
Add default buildSystemRootDir value if system variable does not present

EDbarvinsky:


gradlew.bat
```diff
@@ -2,40 +2,35 @@
...
+
+:fail
 echo.
-echo ERROR: Couldn't find %GRADLEW_PATH%
-echo Check BUILD_SYSTEM_ROOT and buildSystemVersion set correctly
+echo ERROR: %ERROR_MSG%
 echo.
 echo Please check Build System guide:
-echo https://github.com/vividus-framework/vividus-build-system
+echo https://Vividus-BuildSystem
```
valfirst: wrong change, looks like this is a merge error

PR #17, opened 2019-08-18
Add DEBUG to the http methods

Yauhenda:


vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.apache.http.client.methods.HttpRequestBase;
+
+public class HttpDebug extends HttpRequestBase
+{
+    public final static String METHOD_NAME = "DEBUG";
+
+    public HttpDebug()
+    {
+        super();
```
uarlouski: why do you need explicit invocations of no-arg constructor?

valfirst: to have the ability to set `URI` via setter

uarlouski: I was talking about super() invocations, because no-args constructors are invoked in java explicitly before all instructions if we don't specify other constructor

vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+        setURI(uri);
+    }
+
+    public HttpDebug(final String uri)
+    {
+        setURI(URI.create(uri));
+    }
+
+    @SuppressWarnings("checkstyle:SimpleAccessorNameNotation")
```
uarlouski: will this warning occur if we rename METHOD_NAME to METHOD?

vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package org.vividus.http;
+
+import java.net.URI;
+
+import org.apache.http.client.methods.HttpRequestBase;
+
+public class HttpDebug extends HttpRequestBase
```
uarlouski: could you please add java doc to understand where this method comes from?

vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.http.client.methods.HttpRequestBase;
+
+/**
+ * HTTP DEBUG method
+ * <p>
+ * When the client tries to automatically attach the debugger in an .NET application, the client sends a HTTP
+ * request that contains the DEBUG verb. This HTTP request is used to verify that the process of the application is
+ * running and to select the correct process to attach.
+ */
+
```
valfirst: useless empty line

vividus-tests/src/main/resources/story/integration/HttpRequest.story
```diff
@@ -0,0 +1,8 @@
+Description: Integration tests for ApiSteps functionality
+
+Meta:
+    @group vividus-plugin-rest-api
+
+Scenario: Verify method DEBUG is supported
+When I issue a HTTP DEBUG request for a resource with the URL 'http://example.org/'
+Then `${responseCode}` matches `40[35]`
```
valfirst:
```
Then `${responseCode}` is equal to `405`
```
?

vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
```diff
@@ -0,0 +1,54 @@
...
+    public HttpDebug(final String uri)
+    {
+        setURI(URI.create(uri));
+    }
+
+    @Override
+    public String getMethod()
+    {
+        return METHOD;
```
uarlouski: why not to return "DEBUG"?

vividus-tests/src/main/resources/story/integration/HttpRequest.story
```diff
@@ -0,0 +1,8 @@
+Description: Integration tests for ApiSteps functionality
+
+Meta:
+    @group vividus-plugin-rest-api
+
+Scenario: Verify method DEBUG is supported
+When I issue a HTTP DEBUG request for a resource with the URL 'http://example.org/'
+Then `${responseCode}` is equal `405`
```
valfirst: ` is equal to`

vividus-http-client/src/main/java/org/vividus/http/HttpDebug.java
valfirst: because public constant may be reused

PR #23, opened 2019-08-19
Fix image tooltip pattern

EDbarvinsky:


vividus-plugin-web-app/src/main/java/org/vividus/bdd/steps/ui/web/ElementPattern.java
```diff
@@ -27,6 +27,7 @@
     public static final String SELECT_PATTERN = ".//select[@*=%s]";
     public static final String LABEL_PATTERN = ".//label[text()=%1$s or *=%1$s or @*=%1$s]";
     public static final String RADIO_OPTION_INPUT_PATTERN = ".//input[@type='radio' and @id=%s]";
+    public static final String IMAGE_TOOLTIP_PATTERN = ".//img[@alt=%1$s or @title=%1$s]";
```
valfirst: this is considered as anti-pattern, move this constant to the corresponding class or even inline it if it's used once

PR #35, opened 2019-08-21
Drop usage of PowerMock in S3BucketStepsTests

valfirst:


vividus-plugin-aws-s3/src/test/java/org/vividus/aws/s3/steps/S3BucketStepsTests.java
```diff
@@ -64,50 +60,47 @@
...
+        verify(amazonS3Client).putObject(eq(S3_BUCKET_NAME), eq(S3_OBJECT_KEY), argThat(bais -> {
+            try
+            {
+                byte[] actual = IOUtils.toByteArray(bais);
+                bais.reset();
+                return Arrays.equals(csv, actual);
+            }
+            catch (IOException e)
+            {
+                return false;
```
uarlouski: why not to re-throw it? if the exception really occurs it would be hard to understand what is the reason the matching fails

valfirst: in current case the exception will never be thrown, but I need to catch checked exceptions

PR #62, opened 2019-09-04
Update spring configuration

TatsianaLuchonak:


vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/DiffDateExpressionProcessor.java
```diff
@@ -23,9 +23,12 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import javax.inject.Named;
+
 import org.apache.commons.lang3.StringUtils;
 import org.vividus.util.DateUtils;
 
+@Named("diffDateToExpression")
```
valfirst: actual name is not required: `@Named("diffDateToExpression")` -> `@Named`
here and below

vividus/src/main/resources/org/vividus/spring.xml
```diff
@@ -18,6 +18,9 @@
     <context:component-scan base-package="org.vividus.jackson.databind" >
         <context:exclude-filter type="regex" expression="org\.vividus\.jackson\.databind\..*\..*"/>
     </context:component-scan>
+    <context:component-scan base-package="org.vividus.bdd.expression">
+        <context:exclude-filter type="regex" expression="org\.vividus\.bdd\.expression\.DateExpressionProcessor*"/>
```
valfirst: why is it required to exclude it explicitly?

PR #78, opened 2019-09-12
Add tests for BddScenariosCounter

ivan-lednev:


vividus/src/test/java/org/vividus/runner/BddScenariosCounterTests.java
```diff
@@ -0,0 +1,155 @@
...
+        root.addChild(beforeStories);
+        root.addChild(story);
+        story.addChild(beforeStory);
+        story.addChild(scenario);
+
+        when(runner.getDescription()).thenReturn(root);
+
+        BddScenariosCounter.main(new String[0]);
+
+        assertTrue(getOutput().contains(String.format(FORMATTER, 1, STORIES)));
```
valfirst: use `assertThat` + corresponding matcher, it makes assertions more readable

valfirst: here and below in similar cases

valfirst: `String.format(FORMATTER, 1, STORIES)` -> replace with a plain string:
- there is no much sense to use `String.format` with constants only
- plain strings make unit tests more readable

vividus/src/test/java/org/vividus/runner/BddScenariosCounterTests.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test
+    public void testHelpOptionIsPresent() throws Exception
+    {
+        BddScenariosCounter.main(new String[] {"--help"});
+        assertTrue(getOutput().contains("usage: BddScenariosCounter"));
+    }
+
+    private void stubPropertiesAndRunnerInvocations() throws Exception
```
valfirst: https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs

valfirst: `...And...` in the method name usually means that the method does more than one thing, but more than one thing is a violation of [Single responsibility principle](https://stackoverflow.com/questions/46541197/does-the-single-responsibility-principle-work-in-oop/46556381#46556381).  So what can be done here:
- `when(BeanFactory...` can't be used without `PowerMockito.mockStatic(BeanFactory.class);` so they can be grouped and placed into a new method
- mock `runner` is not used in all tests, so it makes sense to move its creation and `whenNew(JUnitReportingRunner...` to another method and this method should return just created mock:
```java
private JUnitReportingRunner mockJUnitReportingRunnerInstantiation()
{
    JUnitReportingRunner runner = mock(JUnitReportingRunner.class);
    whenNew(JUnitReportingRunner.class).withArguments(StoriesRunner.class).thenReturn(runner);
    return runner;
}
```
```

File: vividus/src/test/java/org/vividus/runner/BddStepsCounterTests.java [hunk 3]
```diff
@@ -52,6 +53,7 @@
 
 @RunWith(PowerMockRunner.class)
 @PrepareForTest({ Vividus.class, BeanFactory.class })
+@PowerMockIgnore({ "javax.*", "com.sun.*", "jdk.*", "org.xml.*", "org.w3c.*" })
```
valfirst: this change should go as a separate commit

vividus/src/test/java/org/vividus/runner/BddScenariosCounterTests.java
```diff
@@ -0,0 +1,157 @@
...
+    private static final String STORY = "story description";
+    private static final String SCENARIO = "scenario description";
+    private static final String EXAMPLE = "example description";
+    private static final String BEFORE = "before description";
+    private static final String RESOURCE_LOCATION = "bdd.story-loader.batch1.resource-location";
+
+    @Before
+    public void before()
+    {
+        PowerMockito.mockStatic(Vividus.class);
```
valfirst: is it required?

ivan-lednev: Without it, I get this exception
`java.lang.IllegalStateException: environments is not set`

PR #113, opened 2019-10-08
Additional check for element on visibility

AlenaBudevich:


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/AbstractElementSearchAction.java
```diff
@@ -90,7 +94,9 @@ private boolean isElementVisible(WebElement element, boolean scrolled)
     {
         if (!element.isDisplayed())
         {
-            if (!scrolled)
+            String cssSelector = cssSelectorFactory.getCssSelector(element);
+            boolean isVisible = javascriptActions.executeScript(String.format(VISIBILITY_SCRIPT, cssSelector));
```
uarlouski: 1) what if we pass element as an arg to script and then invoke getBoundingClientRect().... instead of getting css selector from it?
2) isVisible -> visible

PR #126, opened 2019-10-21
[plugin-web-app] Hide scrollbars for headless Chrome browser

valfirst:


vividus-plugin-web-app/src/main/resources/properties/profile/web/headless/chrome/profile.properties
```diff
@@ -1,2 +1,2 @@
 selenium.browser=chrome
-web.driver.CHROME.command-line-arguments=headless
+web.driver.CHROME.command-line-arguments=--headless --hide-scrollbars
```
ikalinin1: Why it is needed? :)

valfirst: to avoid useless multiple scrollbars on full-page screenshots

PR #131, opened 2019-10-24
Set parameter converters before story parser

ivan-lednev:


vividus-tests/src/main/resources/story/integration/TableTransformers.story
```diff
@@ -31,3 +31,13 @@ Examples:
 |column1|column2|
 |A      |C      |
 |B      |D      |
+
+
+Scenario: Update a value through ITERATING with a global variable
+When I initialize the STORY variable `updatedNumber` with value `#{eval(${updatedNumber} + 1)}`
```
ikalinin1: no reason to make updatedNumber global?

PR #133, opened 2019-10-24
[plugin-web-app] Extract device-pixel-ratio automatically

### LINKED ISSUE #71: Screenshooting mechanism works wrong on MacOS
1. Run Visual check OR force test failure on MacOS
2. Open Allure report -> find a screenshot

Result: Only part of the screen is shown at the screenshot

Expected Result: Entire screen on desired Context should be shown. Please see attached Allure report
[allure.zip](https://github.com/vividus-framework/vividus/files/3591279/allure.zip)

valfirst:
Fixes #71

### DIFF
--- a/vividus-plugin-web-app/src/main/java/org/vividus/selenium/screenshot/AshotFactory.java
+++ b/vividus-plugin-web-app/src/main/java/org/vividus/selenium/screenshot/AshotFactory.java
@@ -19,13 +19,17 @@ package org.vividus.selenium.screenshot;
 import static ru.yandex.qatools.ashot.shooting.ShootingStrategies.cutting;
 
 import java.util.Optional;
+import java.util.function.Supplier;
 
 import javax.inject.Inject;
+
+import com.google.common.base.Suppliers;
 
 import org.openqa.selenium.ScreenOrientation;
 import org.vividus.selenium.IWebDriverFactory;
 import org.vividus.selenium.SauceLabsCapabilityType;
 import org.vividus.selenium.manager.IWebDriverManager;
+import org.vividus.ui.web.IRuntimeEnvironment;
 import org.vividus.ui.web.action.IJavascriptActions;
 
 import ru.yandex.qatools.ashot.AShot;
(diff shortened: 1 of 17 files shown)


vividus-plugin-web-app/src/test/java/org/vividus/ui/web/RuntimeEnvironmentTests.java
```diff
@@ -17,38 +17,55 @@
...
     }
 
     @Test
-    void testGetUserAgentMultipleCalls()
+    void testGetUserAgentMultipleTimes()
     {
         runtimeConfiguration.getUserAgent();
         runtimeConfiguration.getUserAgent();
-        Mockito.verify(jsActions).getUserAgent();
+        verify(javascriptActions, times(1)).getUserAgent();
```
uarlouski: times(1) - no need?

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/RuntimeEnvironment.java
```diff
@@ -16,19 +16,30 @@
...
 package org.vividus.ui.web;
 
-import javax.inject.Inject;
-
 import org.vividus.ui.web.action.IJavascriptActions;
 
 public class RuntimeEnvironment implements IRuntimeEnvironment
 {
-    @Inject private IJavascriptActions javascriptActions;
+    private IJavascriptActions javascriptActions;
```
uarlouski: might be final now

vividus-plugin-web-app/src/main/java/org/vividus/selenium/screenshot/AshotFactory.java
```diff
@@ -37,37 +41,35 @@
...
     @Inject private IJavascriptActions javascriptActions;
+    @Inject private IRuntimeEnvironment runtimeEnvironment;
     @Inject private ScreenshotDebugger screenshotDebugger;
 
-    public void init()
-    {
-        baseShootingStrategy = ShootingStrategies.scaling((float) webDriverManager.getDevicePixelRatio());
-    }
+    private Supplier<ShootingStrategy> baseShootingStrategy = Suppliers.memoize(
+        () -> ShootingStrategies.scaling((float) runtimeEnvironment.getDevicePixelRatio()));
```
uarlouski: is there any chance that we get bad scaling results because we convert double-precision value (as DPR stored and returned by invocation window.devicepixelratio) to single precision value?

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/RuntimeEnvironment.java
valfirst: it can't be since it's used in `ThreadLocal` fields initialization

vividus-plugin-web-app/src/test/java/org/vividus/ui/web/RuntimeEnvironmentTests.java
valfirst: I'd like to keep it explicit to emphasize that exactly one invocation is expected, that's important in this test

vividus-plugin-web-app/src/main/java/org/vividus/selenium/screenshot/AshotFactory.java
valfirst: I've tested and haven't faced with such problem, even more for now there is no case when DPR is actually floating point number, it's always integer

ikalinin1: By the way, looks like it's can be floating point https://www.mydevice.io/

PR #137, opened 2019-10-30
[vividus] Add step module name to the output of BddStepPrinter

valfirst:


vividus/src/test/java/org/vividus/runner/BddStepPrinterTests.java
```diff
@@ -97,9 +97,9 @@ public void testPrintToFile() throws Exception
         StepCandidate stepCandidate2 = mockStepCandidate("When", "I do '$action'", "deprecatedMethod");
         StepCandidate stepCandidate3 = mockStepCandidate("Then", "I perform '$verification'", (Method) null);
         when(candidateSteps.listCandidates()).thenReturn(List.of(stepCandidate1, stepCandidate2, stepCandidate3));
-        return List.of("                        Given initial state is '$status'",
-                             "COMPOSITE IN STEPS FILE Then I perform '$verification'",
-                             "DEPRECATED              When I do '$action'");
+        return List.of("                         Given initial state is '$status'",
```
ikalinin1: Could you please cover the logic related to the location?

ikalinin1: Could you please cover the logic related to the location?

vividus/src/main/java/org/vividus/runner/BddStepPrinter.java
```diff
@@ -94,77 +105,105 @@ public static void main(String[] args) throws ParseException, IOException
...
+        if (!compositeInStepsFile)
+        {
+            step.setDeprecated(method.isAnnotationPresent(Deprecated.class));
+            String baseFileName = FilenameUtils.getBaseName(
+                    method.getDeclaringClass().getProtectionDomain().getCodeSource().getLocation().toString());
+            step.setLocation(baseFileName.replaceAll("-\\d+\\.\\d+\\.\\d+.*", ""));
+        }
+        else
+        {
+            step.setLocation("");
```
uarlouski: empty used two times in the file on lines 118 and 122, shouldn't pmd(or checkstyle I don't remebmer :)) fail the build

vividus/src/main/java/org/vividus/runner/BddStepPrinter.java
```diff
@@ -68,15 +71,23 @@ public static void main(String[] args) throws ParseException, IOException
         Vividus.init();
         Set<Step> steps = getSteps();
 
+        int maxLocationLength = steps.stream().map(Step::getLocation).mapToInt(String::length).max().orElse(0);
+
+        List<String> stepsLines = steps.stream().map(s -> String
+                .format("%-" + (maxLocationLength + 1) + "s%-24s%s %s", s.getLocation(),
```
uarlouski: why not %d and then pass (maxLocationLength + 1) as param?

vividus/src/main/java/org/vividus/runner/BddStepPrinter.java
```diff
@@ -94,77 +105,105 @@ public static void main(String[] args) throws ParseException, IOException
...
+            if (this == o)
             {
                 return true;
             }
-            if (obj == null)
-            {
-                return false;
-            }
-            if (obj.getClass() != this.getClass())
+            if (o == null || getClass() != o.getClass())
```
uarlouski: might be replaced with a single instanceof (https://stackoverflow.com/a/26514984)

vividus/src/test/java/org/vividus/runner/BddStepPrinterTests.java
valfirst: I'm doubtful it's really worth: module names are collected only for steps from jar-s, so I need to create a sample jar with steps, publish it somewhere and use it in unit tests, since this functionality is not crucial, I would leave it as is for now

vividus/src/main/java/org/vividus/runner/BddStepPrinter.java
valfirst: do you propose to use `String.format` twice?

vividus/src/main/java/org/vividus/runner/BddStepPrinter.java
valfirst: 1) `instanceof` accepts subclasses, but it's not desired behavior
2) the provided benchmark looks fragile, +0.22% degradation could be just a deviation, and also we use JDK 11 now

Issue #150, opened 2019-11-11
Add possibility to check which cookies are saved


KatsiarynaM:
Please, add possibility to check which cookies are saved by step "When I set browser cookies to the API context"

PR #151, opened 2019-11-11
Add publishing of cookies received from browser

uarlouski:
Screen Shot 2019-11-11 at 8 21 51 PM

Header has its own background since it's sticky as we scroll down

### DIFF
--- a/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/bdd/steps/integration/HttpRequestSteps.java
+++ b/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/bdd/steps/integration/HttpRequestSteps.java
@@ -16,23 +16,27 @@
 
 package org.vividus.bdd.steps.integration;
 
-import java.util.stream.Collectors;
-
-import javax.inject.Inject;
+import java.util.Map;
 
 import org.apache.http.client.CookieStore;
-import org.apache.http.cookie.Cookie;
 import org.jbehave.core.annotations.When;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.vividus.http.HttpTestContext;
+import org.vividus.reporter.event.AttachmentPublisher;
 import org.vividus.ui.web.action.ICookieManager;
 
 public class HttpRequestSteps
 {
-    private static final Logger LOGGER = LoggerFactory.getLogger(HttpRequestSteps.class);
-    @Inject private ICookieManager cookieManager;
-    @Inject private HttpTestContext httpTestContext;
+    private final ICookieManager cookieManager;
+    private final HttpTestContext httpTestContext;
+    private final AttachmentPublisher attachmentPublisher;
+
+    public HttpRequestSteps(ICookieManager cookieManager, HttpTestContext httpTestContext,
+            AttachmentPublisher attachmentPublisher)
+    {
+        this.cookieManager = cookieManager;
+        this.httpTestContext = httpTestContext;
+        this.attachmentPublisher = attachmentPublisher;
+    }
 
     /**
      * Step implemented to reuse browser cookies for HTTP client executed requests.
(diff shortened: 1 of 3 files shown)


vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/bdd/steps/integration/HttpRequestSteps.java
```diff
@@ -16,23 +16,21 @@
...
 import org.vividus.http.HttpTestContext;
+import org.vividus.reporter.event.AttachmentPublisher;
 import org.vividus.ui.web.action.ICookieManager;
 
 public class HttpRequestSteps
 {
-    private static final Logger LOGGER = LoggerFactory.getLogger(HttpRequestSteps.class);
     @Inject private ICookieManager cookieManager;
     @Inject private HttpTestContext httpTestContext;
+    @Inject private AttachmentPublisher attachmentPublisher;
```
valfirst: what about changing to constructor?

valfirst:
Would you mind publishing a screenshot of sample attachment?

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/bdd/steps/integration/HttpRequestSteps.java
uarlouski: are we going to completely migrate from field injection to constructor injection?

valfirst: yes, I think

ikalinin1:
Keep the log, to have an ability triage results without report?

uarlouski:
> Keep the log, to have an ability triage results without report?
1) Ability triage results without report becomes useless when the number of threads grows
2) Do you propose to change logging level to info? otherwise it will not be used at all since end users are not supposed to look into the code and find logging, additionally we have to describe how to change logging level in vividus to make it able for end users to display that info
3) If so, the allure will contain html report and same values from logger, isn't it look redundant?

ikalinin1:
> > Keep the log, to have an ability triage results without report?
> 
> 1. Ability triage results without report becomes useless when the number of threads grows
> 2. Do you propose to change logging level to info? otherwise it will not be used at all since end users are not supposed to look into the code and find logging, additionally we have to describe how to change logging level in vividus to make it able for end users to display that info
> 3. If so, the allure will contain html report and same values from logger, isn't it look redundant?

Debug information still could useful.

uarlouski:
> > > Keep the log, to have an ability triage results without report?
> > 
> > 
> > 
> > 1. Ability triage results without report becomes useless when the number of threads grows
> > 2. Do you propose to change logging level to info? otherwise it will not be used at all since end users are not supposed to look into the code and find logging, additionally we have to describe how to change logging level in vividus to make it able for end users to display that info
> > 3. If so, the allure will contain html report and same values from logger, isn't it look redundant?
> 
> Debug information still could useful.

This debug info was never displayed and the end users never knew and will never know about these logs, so you purpose to change debug level to info/increase logging level?

ikalinin1:
> > > > Keep the log, to have an ability triage results without report?
> > > 
> > > 
> > > 
> > > 1. Ability triage results without report becomes useless when the number of threads grows
> > > 2. Do you propose to change logging level to info? otherwise it will not be used at all since end users are not supposed to look into the code and find logging, additionally we have to describe how to change logging level in vividus to make it able for end users to display that info
> > > 3. If so, the allure will contain html report and same values from logger, isn't it look redundant?
> > 
> > 
> > Debug information still could useful.
> 
> This debug info was never displayed and the end users never knew and will never know about these logs, so you propose to change debug level to info/increase logging level and at the same time keep the data twice in logs and in HTML report?

Yep, also we should revise logging, since we have lots of debug level logging;
https://github.com/vividus-framework/vividus/search?p=1&q=%22LOGGER.debug%22&unscoped_q=%22LOGGER.debug%22

valfirst:
Closes #150

PR #171, opened 2019-11-18
[http-client] Make HttpRequestBuildException checked

valfirst:


vividus-http-client/src/main/java/org/vividus/http/HttpRequestBuilder.java
```diff
@@ -86,7 +86,7 @@ public HttpRequestBuilder withHeaders(List<Header> headers)
         return this;
     }
 
-    public HttpRequestBase build()
```
ikalinin1: could you please clarify a reason behind the changes?

valfirst: https://github.com/vividus-framework/vividus/blob/master/vividus-plugin-rest-api/src/main/java/org/vividus/http/HttpRequestExecutor.java#L103 - the exception is caught here, but it's not thrown here, it"s thrown in some wrapped methods, it's really hard to trace unchecked exception with such approach

PR #173, opened 2019-11-20
Add bug report templates

valfirst:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+**To Reproduce**
+Steps to reproduce the behavior.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
```
uarlouski: does it make sense to add point about API bugs?

ikalinin1: I think we should also request environment info such as: 
- Vividus version
- OS/version where tests were running
- Java version
- Logs

valfirst: which sections do you propose to add?

uarlouski: never mind, all I wanted to say is spread across other points

PR #180, opened 2019-11-21
Handle exception for nonexistent cell in excel file

AlenaBudevich:


vividus-plugin-excel/src/main/java/org/vividus/excel/ExcelSheetParser.java
```diff
@@ -219,6 +219,10 @@ public String getDataFromCell(String address)
     {
         CellReference cellReference = new CellReference(address);
         Row row = sheet.getRow(cellReference.getRow());
+        if (row == null)
+        {
+            throw new IllegalArgumentException(String.format("Cell by address '%s' doesn't exist", address));
```
valfirst: `Row at address ...`

PR #197, opened 2019-11-30
Fix cookie template to handle empty expire date

### LINKED ISSUE #195: Cookies errors are not identified correctly in Allure report
Executing a step "When I set browser cookies to the API context" leads to an error 

> Error executing FreeMarker template freemarker.core.InvalidReferenceException: The following has evaluated to null or missing: ==> cookie.expiryDate [in template "org/vividus/bdd/steps/integration/browser-cookies.ftl" at line 72, column 27]
The step is not marked as failed in Allure report, it is displayed when the step details are expanded.

Steps to reproduce:
1. Execute:
Scenario: Test cookies
Given I am on a page with the URL 'https://www.epam.com/'
When I set browser cookies to the API context
2. Check the Allure report
**Expected:** the error is identified correctly
**Actual:** Incorrect identification
image

See Allure report:
[allure.zip](https://github.com/vividus-framework/vividus/files/3905536/allure.zip)

uarlouski:
Closes #195

### DIFF
--- a/vividus-plugin-web-app-to-rest-api/src/main/resources/org/vividus/bdd/steps/integration/browser-cookies.ftl
+++ b/vividus-plugin-web-app-to-rest-api/src/main/resources/org/vividus/bdd/steps/integration/browser-cookies.ftl
@@ -69,7 +69,11 @@
                         ${cookie.path}
                     </td>
                     <td>
-                        ${cookie.expiryDate?datetime}
+                        <#if cookie.expiryDate??>
+                            ${cookie.expiryDate?datetime}
+                        <#else>
+                            Session
+                        </#if>
                     </td>
                     <td>
                         ${cookie.persistent?c}


vividus-plugin-web-app-to-rest-api/src/main/resources/org/vividus/bdd/steps/integration/browser-cookies.ftl
```diff
@@ -69,7 +69,9 @@
                         ${cookie.path}
                     </td>
                     <td>
-                        ${cookie.expiryDate?datetime}
+                        <#if cookie.expiryDate??>
+                            ${cookie.expiryDate?datetime}
+                        </#if>
```
valfirst: Shouldn't you put some value if no expiry date set? What does Chrome show?

uarlouski: Good catch, as rfc says If a cookie has neither the Max-Age nor the Expires attribute, the user agent will retain the cookie until "the current session is over" (as defined by the user agent). Took default value from chrome

PR #207, opened 2019-12-06
Fix KnownIssueDataProvider returns null

Yauhenda:


vividus-plugin-web-app/src/main/java/org/vividus/bdd/issue/CurrentPageUrlKnownIssueDataProvider.java
```diff
@@ -21,13 +21,15 @@
...
 public class CurrentPageUrlKnownIssueDataProvider implements IKnownIssueDataProvider
 {
     @Inject private IWebDriverProvider webDriverProvider;
 
     @Override
-    public String getData()
+    public Optional<String> getData()
     {
-        return webDriverProvider.get().getCurrentUrl();
+        return Optional.ofNullable(webDriverProvider.get().getCurrentUrl());
```
valfirst: I believe it won't return `null`, it will throw `WebDriverException` which should be caught

vividus-soft-assert/src/main/java/org/vividus/softassert/issue/KnownIssueChecker.java
```diff
@@ -123,7 +123,8 @@ private boolean doAllDynamicPatternsMatch(Map<String, Pattern> dynamicPatterns)
                 String key = entry.getKey();
                 Pattern pattern = entry.getValue();
                 return Optional.ofNullable(knownIssueDataProviders.get(key))
-                           .map(p -> pattern.matcher(p.getData()).matches())
+                        .flatMap(IKnownIssueDataProvider::getData)
```
valfirst: new unit test is required to cover this logic

vividus-plugin-web-app/src/main/java/org/vividus/bdd/issue/CurrentPageUrlKnownIssueDataProvider.java
```diff
@@ -26,8 +29,15 @@
...
+    public Optional<String> getData()
     {
-        return webDriverProvider.get().getCurrentUrl();
+        try
+        {
+            return Optional.ofNullable(webDriverProvider.get().getCurrentUrl());
+        }
+        catch (WebDriverException e)
+        {
+            return Optional.empty();
```
valfirst: unit test is needed to cover this code branch

PR #264, opened 2020-01-17
Add default variables

ikalinin1:


vividus-tests/src/main/resources/story/integration/VariableTests.story
```diff
@@ -0,0 +1,10 @@
+Description: Integration tests for variables
+
+Meta:
+    @group variables
+
+Scenario: Verify default variables
+Then `1` is = `${key:1}`
+Then `2` is = `${list[0]:2}`
+Then `3` is = `${map.key:4}`
```
valfirst: why should it be equal to `3`?

ikalinin1: unsaved file, sorry about that

PR #272, opened 2020-01-21
[plugin-rest-api] Add generic step to wait for JSON element presence

valfirst:
The step exeutes sub-steps and waits for a specified amount of time
 until HTTP response body contains an element by the specified JSON path.

### DIFF
--- a/vividus-util/src/main/java/org/vividus/util/function/CheckedRunnable.java
+++ b/vividus-util/src/main/java/org/vividus/util/function/CheckedRunnable.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2019-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


vividus-util/src/main/java/org/vividus/util/function/CheckedRunnable.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.vividus.util.function;
+
+@FunctionalInterface
+public interface CheckedRunnable<E extends Exception>
```
uarlouski: maybe Exception  -> Throwable to cover all possible cases?

valfirst: "checked" here means checked exception-s only

uarlouski: Throwable is also checked by its nature, if you meant that by doing this we would be able to declare also Error and RE exception then I think such cases should be prohibited by static code analysis tools

valfirst: Please propose the rule for static code analysis :bulb:

uarlouski: there is such rule in PMD https://github.com/pmd/pmd/issues/1692

valfirst: this rule about method declaration, how to apply it to this case?

uarlouski: mixed this up :)

PR #311, opened 2020-02-11
Add ability to scroll context element horizontally

### LINKED ISSUE #140: The framework hasn't step for horizontal page scrolling
Problem: the framework hasn't step for horizontal page scrolling.

Our application requires horizontal scrolling of the page, so as the user scrolls the page horizontally, the data continues to be loaded. Accordingly, to make sure that all the data has arrived and there are no differences between the attributes (columns) in the database and the application, you need to scroll horizontally on the page, only after that you can clearly judge that all the data has arrived and the page is loaded without visible modal windows with errors.
In our application, the scroll looks like this:
2019-11-01 14_43_26-Window

Could you please add this step in Framework,
Kindly regards

### LINKED ISSUE #260: Could you please add an integration test for element scroll
Could you please add an integration test

_Originally posted by @ikalinin1 in https://github.com/vividus-framework/vividus/pull/259/files_

ikalinin1:
Closes #140
Closes #260

### DIFF
--- a/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/IJavascriptActions.java
+++ b/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/IJavascriptActions.java
@@ -63,6 +63,10 @@ public interface IJavascriptActions
 
     void scrollToStartOf(WebElement element);
 
+    void scrollToLeftOf(WebElement element);
+
+    void scrollToRightOf(WebElement element);
+
     /**
      * Opens page URL in a new window
      * @param pageUrl An absolute URL of the page
--- a/vividus-plugin-web-app/src/main/resources/steps/defaults/scroll.steps
+++ b/vividus-plugin-web-app/src/main/resources/steps/defaults/scroll.steps
@@ -1,0 +1,5 @@
+Composite: When I scroll to the end of the context
+When I scroll context to BOTTOM edge
+
+Composite: When I scroll to the start of the context
+When I scroll context to TOP edge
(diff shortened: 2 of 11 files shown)


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/IJavascriptActions.java
```diff
@@ -63,6 +63,10 @@
 
     void scrollToStartOf(WebElement element);
 
+    void scrollHorizontallyToStartOf(WebElement element);
```
valfirst: `scrollToLeftOf` ? to keep the same naming scheme

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/IJavascriptActions.java
```diff
@@ -63,6 +63,10 @@
...
+
+    void scrollHorizontallyToEndOf(WebElement element);
```
valfirst: `scrollToRightOf`

vividus-plugin-web-app/src/main/resources/steps/defaults/scroll.steps
```diff
@@ -0,0 +1,11 @@
+Composite: When I scroll to the end of the page
+When I scroll context to BOTTOM edge
```
valfirst: risky change, there could be cases, when some context is set and user wants to scroll the page, but not context

vividus-plugin-web-app/src/main/resources/steps/defaults/scroll.steps
```diff
@@ -0,0 +1,11 @@
...
+
+Composite: When I scroll to the start of the page
+When I scroll context to TOP edge
```
valfirst: risky change, there could be cases, when some context is set and user wants to scroll the page, but not context

vividus-plugin-web-app/src/main/resources/steps/defaults/scroll.steps
ikalinin1: It is, but shouldn't we cut it as soon as possible?

valfirst: breaking changes are always painful for users, so I recommend to Deprecate these steps

PR #338, opened 2020-02-22
Use org.jbehave.core.configuration.Configuration as a base 

ikalinin1:
to avoid early initialization of configuration


vividus-bdd-engine/src/main/java/org/vividus/bdd/spring/Configuration.java
```diff
@@ -41,8 +40,11 @@
 import org.vividus.bdd.steps.ParameterAdaptor;
 import org.vividus.bdd.steps.ParameterConvertersDecorator;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 @SuppressWarnings("rawtypes")
-public class Configuration extends MostUsefulConfiguration
+@SuppressFBWarnings("NM_SAME_SIMPLE_NAME_AS_SUPERCLASS")
+public class Configuration extends org.jbehave.core.configuration.Configuration
```
valfirst: what about renaming to `ExtendedConfiguration` ?

PR #353, opened 2020-02-28
Add ability to skip base story if given story is failed

valfirst:


vividus/src/main/resources/properties/defaults/default.properties
```diff
@@ -23,6 +23,7 @@ bdd.configuration.composite-paths=
 bdd.configuration.formats=
 bdd.configuration.dry-run=false
 bdd.configuration.reset-state-before-scenario=true
+bdd.configuration.skip-story-if-given-story-failed=false
```
ikalinin1: Please add integration tests

PR #360, opened 2020-03-01
[tests] Add integration tests for GivenStories

valfirst:


vividus-tests/src/main/resources/story/integration/Use GivenStories.story
```diff
@@ -0,0 +1,11 @@
+Meta:
+    @group bdd-engine
+
+GivenStories: /story/integration/Precondition.story
+
+Scenario: This scenario should run after story-level GivenStories
+Then `base-true` is equal to `base-true`
+
+Scenario: This scenario should run story-level and sceario-level GivenStories
```
anastasi-n: sceNario-level

vividus-tests/src/main/resources/story/integration/This Story should not run because GivenStory is failed.story
```diff
@@ -0,0 +1,7 @@
+Meta:
+    @group bdd-engine
+
+GivenStories: /story/integration/KnownIssues.story
+
+Scenario: This scenario should never run, becuase GivenStory is failed and `bdd.configuration.skip-story-if-given-story-failed=true`
```
anastasi-n: becAuse

PR #367, opened 2020-03-05
Fix resolving of variables with special chars in names

valfirst:


vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java
```diff
@@ -112,26 +112,29 @@ public void clearVariables()
     @SuppressWarnings("unchecked")
     private Object getVariable(Map<String, Object> variables, String key)
     {
-        Matcher variableMatcher = VARIABLE_PATTERN.matcher(key);
-        if (!variableMatcher.find())
+        Object variable = variables.get(key);
```
ikalinin1: could it be 
```java
if (variable != null)
{
    return variable;
}
```
to decreas method complexity ?

vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java
```diff
@@ -112,26 +112,29 @@ public void clearVariables()
...
-            }
-            String mapKey = variableMatcher.group(MAP_KEY_GROUP);
-            if (mapKey != null && variable instanceof Map)
+            Matcher variableMatcher = VARIABLE_PATTERN.matcher(key);
+            if (variableMatcher.find())
             {
-                variable = ((Map<String, ?>) variable).get(mapKey);
+                String variableKey = variableMatcher.group(VARIABLE_NAME_GROUP);
+                variable = variables.get(variableKey);
+                if (variable != null)
```
ikalinin1: would be cool to refactor it to something like:

```java
if (variable != null)
{
    variable = resolveAsList(...)
    variable = resolveAsMap(...)
}

File: vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java (hunk 1 above)
Comment (valfirst) [reply vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java]: it doesn't make it better:
```java
        Object variable = variables.get(key);
        if (variable != null)
        {
            return variable;
        }
        Matcher variableMatcher = VARIABLE_PATTERN.matcher(key);
        if (variableMatcher.find())
        {
            String variableKey = variableMatcher.group(VARIABLE_NAME_GROUP);
            variable = variables.get(variableKey);
            if (variable != null)
            {
                String listIndex = variableMatcher.group(LIST_INDEX_GROUP);
                if (listIndex != null)
                {
                    List<?> listVariable = (List<?>) variable;
                    int elementIndex = Integer.parseInt(listIndex);
                    variable = elementIndex < listVariable.size() ? listVariable.get(elementIndex) : null;
                }
                String mapKey = variableMatcher.group(MAP_KEY_GROUP);
                if (mapKey != null && variable instanceof Map)
                {
                    variable = ((Map<String, ?>) variable).get(mapKey);
                }
            }
        }
        return variable;
```

PR #374, opened 2020-03-05
Return human-readable description by ComparisonRule.toString()

### LINKED ISSUE #373: Report comparison rules in plain English
Current comparison rules reported as EQUAL_TO. To increase readability we should convert them to plain text, **equal to**

ikalinin1:
Closes #373

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/ComparisonRule.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/ComparisonRule.java
@@ -94,6 +94,11 @@ public enum ComparisonRule implements IComparisonRule
                 .findFirst();
     }
 
+    public String toString()
+    {
+        return super.toString().replace('_', ' ').toLowerCase();
+    }
+
     public String getSign()
     {
         return sign;
--- a/vividus-bdd-engine/src/test/java/org/vividus/bdd/steps/ComparisonRuleTests.java
+++ b/vividus-bdd-engine/src/test/java/org/vividus/bdd/steps/ComparisonRuleTests.java
@@ -54,6 +54,20 @@ class ComparisonRuleTests
         // @formatter:on
     }
 
+    static Stream<Arguments> getComparisonRuleAsString()
+    {
+        // @formatter:off
+        return Stream.of(
+            Arguments.of("less than", ComparisonRule.LESS_THAN),
+            Arguments.of("less than or equal to", ComparisonRule.LESS_THAN_OR_EQUAL_TO),
+            Arguments.of("greater than", ComparisonRule.GREATER_THAN),
+            Arguments.of("greater than or equal to", ComparisonRule.GREATER_THAN_OR_EQUAL_TO),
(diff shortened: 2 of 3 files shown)


vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/ComparisonRule.java
```diff
@@ -94,6 +96,12 @@ public static ComparisonRule parse(String sign)
                 .findFirst();
     }
 
+    @Override
+    public String toString()
+    {
+        return asString;
```
valfirst:
```java
return name().toLowerCase().replace('_', ' ');
```
?

vividus-plugin-html/src/main/java/org/vividus/bdd/steps/html/HtmlSteps.java
```diff
@@ -56,6 +48,36 @@
...
+     * @see <a href="https://jsoup.org/apidocs/org/jsoup/select/Selector.html"><i>Jsoup Selector API</i></a>
+     * @return true if elements quantity corresponds to rule, false otherwise
+     */
+    @Then("number of elements found by CSS selector `$cssSelector` in HTML `$html` is $comparisonRule `$quantity`")
+    public boolean doesElementByCssSelectorExist(String cssSelector, String html, ComparisonRule comparisonRule,
+            int number)
+    {
+        Elements elements = findElements(html, cssSelector);
+        return softAssert.assertThat(String.format("Number of elements found by CSS selector '%s' is %s %d",
+                cssSelector, comparisonRule, number), elements.size(), comparisonRule.getComparisonRule(number));
```
valfirst: do you really need to add `comparisonRule` and `number` to assertion description? it could be excessive since similar piece of information is added by Matcher

vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/ComparisonRule.java
ikalinin1: I considered an extra code execution for each toString() method call, when values already defined, as an overhead

vividus-plugin-html/src/main/java/org/vividus/bdd/steps/html/HtmlSteps.java
ikalinin1: I think we should pick one solution. BaseValidation uses this approach.
If we will rely on matcher, the PR is obsolete, but we should refactor basevalidations' code

vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/ComparisonRule.java
```diff
@@ -94,6 +94,11 @@ public static ComparisonRule parse(String sign)
                 .findFirst();
     }
 
+    public String toString()
+    {
+        return super.toString().replaceAll("_", " ").toLowerCase();
```
valfirst: since JDK 11 is still our baseline, `.replaceAll("_", " ")` -> `.replace('_', ' ')`

PR #384, opened 2020-03-09
Add ability to configure impersonating mitm

uarlouski:
- proxy.mitm.enabled - enable mitm, default value is false
- proxy.mitm.key-store.alias - alias of certificate in a keystore
- proxy.mitm.key-store.path - path to a keystore
- proxy.mitm.key-store.password - password from a keystore
- proxy.mitm.key-store.type - keystory type


vividus-plugin-web-app/src/main/resources/spring.xml
```diff
@@ -36,8 +36,20 @@
 
     <bean id="proxy" class="org.vividus.proxy.ThreadedProxy" />
 
+    <bean id="mitmManagerFactory" class="org.vividus.proxy.mitm.MitmManagerFactory" />
+
+    <bean id="mitmManagerOptions" class="org.vividus.proxy.mitm.MitmManagerOptions" >
+        <constructor-arg index="0" value="${proxy.mitm.key-store:#{null}}" />
```
ikalinin1: use default values for properties

vividus-http-client/src/main/resources/properties/defaults/run.properties
```diff
@@ -0,0 +1,3 @@
+http.key-store.path=
+http.key-store.password=
+http.key-store.type=
```
valfirst: I think it should be `JKS` as it was in spring config

PR #385, opened 2020-03-11
Add ability to use filtering transformer for tables from external files

AlenaBudevich:


valfirst:
Please add integration tests

vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/FilteringTableTransformer.java
```diff
@@ -37,6 +43,8 @@
     private static final String BY_MAX_ROWS_PROPERTY = "byMaxRows";
     private static final String BY_COLUMNS_NAMES_PROPERTY = "byColumnNames";
 
+    @Inject private ExtendedConfiguration configuration;
```
valfirst: to avoid circular bean references please use this approach to inject `ExamplesTableFactory`: https://github.com/vividus-framework/vividus/blob/8c58ece42b7816e1e4632cedf9f8c9463acc9eff/vividus-plugin-applitools/src/main/resources/spring.xml#L32

PR #391, opened 2020-03-13
Recreate story manager right before batch execution

### LINKED ISSUE #376: Story run fails with an error 'sleep interrupted' on 0.2.1-SNAPSHOT
One and the same test story fails in the last scenario on 0.2.1 with an error: '**java.lang.InterruptedException: sleep interrupted**'. 
On 0.2.0 the story finishes with 100% success.

[log4 (4).log](https://github.com/vividus-framework/vividus/files/4298568/log4.4.log)

[log4.log](https://github.com/vividus-framework/vividus/files/4298624/log4.log)
_Allure reports are available too, but can be shared personally only._

**Execution environment**
 - Vividus version: [0.2.0] and [0.2.1-SNAPSHOT] 
 - Java: [Version 8 Update 231 (build 1.8.0_231-b11)] JDK 11.0.1

**Desktop Web**
 - OS: [Windows 10]
 - Browser [chrome]
 - Version [80.0.3987.122]

**Additional context**

valfirst:
JBehaveJUnitRunner may initialize story manager with default PerformableTree and
EmebedderControls, so story manager should be re-created before actual batch execution

Issues fixed by this change:
 - Fixes #376 (root cause: batch embedder controls are ignored by story manager)
 - Warnings like
    ```
    2020-03-13 18:06:25,876 SpringContextShutdownHook WARN Unable to register Log4j shutdown hook because JVM is shutting down. Using SimpleLogger
    ```
    The root cause of this warning is skipped reporting of `AfterStories` when single batch is used. (Allure report is generated in `AfterStories` by default or in JVM shutdown hook in case of premature end of test execution. When Allure JVM shutdown hook is triggered, Log4j JVM hook is already finished and Log4j is shutdown, as result Allure reporter can't log anything)

### DIFF
--- a/vividus-bdd-engine/src/test/java/org/vividus/bdd/BatchedEmbedderTests.java
+++ b/vividus-bdd-engine/src/test/java/org/vividus/bdd/BatchedEmbedderTests.java
@@ -27,6 +27,7 @@ import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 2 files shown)


vividus-bdd-engine/src/test/java/org/vividus/bdd/BatchedEmbedderTests.java
```diff
@@ -248,9 +244,31 @@ private void mockBatchExecutionConfiguration()
 
     private boolean assertEmbedderControls(EmbedderControls controls)
     {
-        return controls.threads() == THREADS
-                && "3600".equals(controls.storyTimeouts())
-                && controls.ignoreFailureInStories()
-                && !controls.generateViewAfterStories();
+        return controls.threads() == THREADS && "3600".equals(controls.storyTimeouts()) && controls
+                .ignoreFailureInStories() && !controls.generateViewAfterStories();
```
uarlouski: noisy changes

PR #402, opened 2020-03-18
Leave original columns order after work FILTERING transformer

AlenaBudevich:


vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/FilteringTableTransformer.java
```diff
@@ -65,7 +66,7 @@ public String transform(String tableAsString, ExamplesTableProperties properties
                 ? headerValues.size() : Math.min(headerValues.size(), Integer.parseInt(byMaxColumns));
         List<String> filteredColumnsNames = byColumnNames == null ? headerValues.subList(0, columnsLimit)
                 : Arrays.stream(StringUtils.split(byColumnNames, ';')).map(String::trim).collect(Collectors.toList());
-        Set<String> filteredHeaders = new HashSet<>(filteredColumnsNames);
+        Set<String> filteredHeaders = new TreeSet<>(filteredColumnsNames);
```
valfirst: I think, we should keep the same column order as it was before filtering, what about using `LinkedHashSet` ?

vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/FilteringTableTransformer.java
```diff
@@ -98,6 +85,24 @@ public String transform(String tableAsString, ExamplesTableProperties properties
                 .orElseGet(examplesTable::getRows);
     }
 
+    private List<String> getFilteredHeaders(String byMaxColumns, String byColumnNames, List<String> headerValues)
+    {
+        int columnsLimit = byMaxColumns == null
+                ? headerValues.size() : Math.min(headerValues.size(), Integer.parseInt(byMaxColumns));
+        List<String> filteredColumns = headerValues;
```
valfirst: `List<String> filteredColumns = new ArrayList<>(headerValues);` - copy is required to avoid changes in the original input

PR #414, opened 2020-03-21
Fixes

uarlouski:
- Remove unused GENERIC_CAPABILITIES variable from WebDriverFactory
- Add ability to set nested desired capabilities by separating levels of nesting using commas

### DIFF
--- a/vividus-util/src/main/java/org/vividus/util/property/PropertyParser.java
+++ b/vividus-util/src/main/java/org/vividus/util/property/PropertyParser.java
@@ -19,11 +19,15 @@ package org.vividus.util.property;
 import static java.util.stream.Collectors.toMap;
 import static org.apache.commons.lang3.StringUtils.removeStart;
 
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
+
+import org.apache.commons.lang3.StringUtils;
 
 public class PropertyParser implements IPropertyParser
 {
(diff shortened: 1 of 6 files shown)


vividus-util/src/main/java/org/vividus/util/property/PropertyParser.java
```diff
@@ -43,6 +45,43 @@
...
+    {
+        Map<String, Object> container = new HashMap<>();
+        getPropertyValuesByPrefix(propertyPrefix).forEach((k, v) -> putByPath(container, k, v));
+        return container;
+    }
+
+    @SuppressWarnings("unchecked")
+    private static void putByPath(Map<String, Object> container, String path, String value)
+    {
+        String[] paths = path.split("\\.");
```
valfirst: `org.apache.commons.lang3.StringUtils#split(java.lang.String)`

uarlouski: changed to org.apache.commons.lang3.StringUtils#split(java.lang.String,char) ( space is default sep)

vividus-util/src/main/java/org/vividus/util/property/PropertyParser.java
```diff
@@ -43,6 +47,43 @@
...
+                .limit(limit)
+                .reduce(container, (map, pathKey) ->
+                {
+                    Map<String, Object> nested = (Map<String, Object>) map.compute(pathKey, (k, v) ->
+                    {
+                        if (v == null)
+                        {
+                            return new HashMap<>();
+                        }
+                        else if (!(v instanceof Map))
```
ikalinin1:
```java
else if (v instanceof Map)
{
    return v;
}
throw new IllegalArgumentException(String.format(
                                    "Path key '%s' from path '%s' is already used as a property key", pathKey, path));
```

valfirst:
All changes have been cherry-picked to `master` branch

PR #419, opened 2020-03-23
[vividus-tests] Fix Path variable extracting depending from OS

Ulyana-Yeryksonava:


ikalinin1:
Will be resolved in #1046

PR #420, opened 2020-03-23
Remove exclusion of json-smart library

valfirst:
We can't completely exclude `json-smart` library, because it's
explicitly referenced in internal API of `json-path` library:
https://github.com/json-path/JsonPath/blob/master/json-path/src/main/java/com/jayway/jsonpath/internal/filter/ValueNodes.java#L150-L151

The exclusion causes the errors like:
`java.lang.ClassNotFoundException: net.minidev.json.parser.ParseException`


valfirst:
One more similar issue in `json-path` lib: https://github.com/json-path/JsonPath/issues/274

PR #431, opened 2020-03-25
[plugin-web-app] Fix unused capabilities for local driver instance

ikalinin1:


vividus-plugin-web-app/src/main/java/org/vividus/selenium/WebDriverFactory.java
```diff
@@ -116,8 +116,9 @@ public WebDriver getWebDriver(DesiredCapabilities desiredCapabilities)
 
     private WebDriver getWebDriver(WebDriverType webDriverType, DesiredCapabilities desiredCapabilities)
     {
+        DesiredCapabilities localDesiredCapabilities = getWebDriverCapabilities(true, desiredCapabilities);
         WebDriverConfiguration configuration = getWebDriverConfiguration(webDriverType, true);
```
valfirst: please extract `true` in both cases to a local variable for better readability

vividus-plugin-web-app/src/main/java/org/vividus/selenium/WebDriverFactory.java
```diff
@@ -116,8 +116,9 @@ public WebDriver getWebDriver(DesiredCapabilities desiredCapabilities)
 
     private WebDriver getWebDriver(WebDriverType webDriverType, DesiredCapabilities desiredCapabilities)
```
valfirst: it's not really in scope, but what about inlining this method?

PR #442, opened 2020-03-27
[vividus] Add multiprofiles support

uarlouski:


vividus/src/integrationTest/resources/properties/configuration.properties
```diff
@@ -1,2 +1 @@
-configuration.profile=
```
valfirst: why is it needed to be removed?

uarlouski: with this property here we can't check the following cases in integration tests:
- only configuration.profiles is set
- neither configuration.profiles nor configuration.profile is set

PR #476, opened 2020-04-10
[Github Actions] Publish results to SonarQube once

valfirst:
Previously the run for each platform published the results to SonarQube, but it's redundant, since each publishing overrode the results from another platforms. Now the results are published from Ubuntu only.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -55,15 +55,20 @@ jobs:
 
     - name: Build with Gradle
       env:
+        MATRIX_PLATFORM: ${{ matrix.platform }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -55,15 +55,20 @@ jobs:
 
     - name: Build with Gradle
       env:
+        MATRIX_PLATFORM: ${{ matrix.platform }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       shell: bash
       run: |
-        if [[ -n $SONAR_TOKEN ]]; then
+        if [ "$MATRIX_PLATFORM" = "ubuntu-latest" ]; then
```
uarlouski, line 63: please use [[ for consistency

Issue #496, opened 2020-04-21
Implement capability for JSON transformation to a plain structure


remiprink:
In order to perform ETL testing with NoSQL sources, there should be a possibility to transform the result of query execution against MongoDB to plain structure.
Requirements:
- Possibility to choose object names to include into resulting table (e.g. via JSONPath)
- Possibility to save resulting table into any plain structure (e.g. CSV)

valfirst:
@remiprink please take a look: https://github.com/vividus-framework/vividus/pull/498/files

PR #498, opened 2020-04-22
Add example of how to compare data pulled from mongo with data pulled from relational db

uarlouski:
Relates #496


vividus-tests/src/main/resources/story/system/MongodbStepsTests.story
```diff
@@ -119,6 +130,18 @@ Then a JSON element from '${native-find.cursor}' by the JSON path '$.firstBatch'
    }
 ]'
 
+Scenario: Compare document from MongoDB with table from Postgree
```
valfirst, line 133: Postgree -> relational DB

PR #516, opened 2020-04-30
Add caching of system and integration tests history

uarlouski:


.github/workflows/gradle.yml
```diff
@@ -84,9 +84,20 @@ jobs:
       with:
         fail_ci_if_error: true
 
+    - name: Cache Integration tests trends
+      if: github.ref == 'refs/heads/master' && matrix.platform == 'ubuntu-latest' && always()
+      uses: actions/cache@v1
+      with:
+        path: vividus-tests/output/integration-trends
```
valfirst, line 91: what about using single cache for all trends: `vividus-tests/output/history`?
and have a separation inside the directory: 
- `vividus-tests/output/history/integration-tests`
- `vividus-tests/output/history/system-tests`

.github/workflows/gradle.yml
```diff
@@ -84,9 +84,20 @@ jobs:
       with:
         fail_ci_if_error: true
 
+    - name: Cache Integration tests trends
```
valfirst, line 87: it's better to use name `history` instead of `trends`, because there was a request to track test history as well

PR #534, opened 2020-05-05
[GithubActions] Ignore certificate errors for Integration tests

valfirst:
*.w3schools.com certificate is expired, need to ignore the error
to make integration tests passed

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -95,9 +95,10 @@ jobs:
 
     - name: Integration tests
       if: matrix.platform == 'ubuntu-latest'
-      run: | 
+      run: |
         ./gradlew :vividus-tests:runStories -Pvividus.configuration.environments=integration \
-                                            -Pvividus.allure.history-directory=output/history/integration-tests
+                                            -Pvividus.allure.history-directory=output/history/integration-tests \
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -95,9 +95,10 @@ jobs:
 
     - name: Integration tests
       if: matrix.platform == 'ubuntu-latest'
-      run: | 
+      run: |
         ./gradlew :vividus-tests:runStories -Pvividus.configuration.environments=integration \
-                                            -Pvividus.allure.history-directory=output/history/integration-tests
+                                            -Pvividus.allure.history-directory=output/history/integration-tests \
+                                            -Pvividus.web.driver.CHROME.command-line-arguments="--headless --hide-scrollbars --ignore-certificate-errors"
```
uarlouski, line 101: why --headless --hide-scrollbars params are also necessary here?

valfirst: https://github.com/vividus-framework/vividus/blob/master/vividus-tests/src/main/resources/properties/configuration.properties#L1

https://github.com/vividus-framework/vividus/blob/master/vividus-plugin-web-app/src/main/resources/properties/profile/web/headless/chrome/profile.properties#L2

PR #566, opened 2020-05-12
[plugin-web-app] Fix scrolling script

ikalinin1:
Script scrolling element into viewport center was executed endlessly with timeout 500 seconds;
The strategy brokes the tests for example in the case when the user clicked anchor link, the page has never scrolled to the anchor because of the script;
New approach: Add scroll waiting listener -> Wait for the scroll end -> remove listener -> call selenium exit callback

### DIFF
--- a/vividus-plugin-web-app/src/main/resources/org/vividus/ui/web/action/scroll-element-into-viewport-center.js
+++ b/vividus-plugin-web-app/src/main/resources/org/vividus/ui/web/action/scroll-element-into-viewport-center.js
@@ -1,20 +1,32 @@
 var elementToScroll = arguments[0];
 var exit = arguments[arguments.length-1];
-(scrollElementIntoViewportCenter = function() {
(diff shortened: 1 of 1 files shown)


vividus-plugin-web-app/src/main/resources/org/vividus/ui/web/action/scroll-element-into-viewport-center.js
```diff
@@ -1,20 +1,32 @@
...
-}) ();
+    currentWindow.addEventListener('scroll', clearTimeoutAndWait, false);
+    currentWindow.scrollBy(0, elementToScroll.getBoundingClientRect().top - currentWindow.innerHeight * 0.25);
+}
+catch(e) {
+    // swallow error quietly
+}
+
+function wait() {
+waitForScroll =  setTimeout(function() {
```
valfirst, line 21: please indent the line

PR #572, opened 2020-05-13
Refactor frame interactions

ikalinin1:


vividus-plugin-web-app/src/main/java/org/vividus/bdd/steps/ui/web/SetContextSteps.java
```diff
@@ -240,7 +158,7 @@ public void switchingToDefault()
     }
 
     /**
-     * Switching to the frame with xpath
+     * Switching to the frame using on of supported locators
```
uarlouski, line 161: on -> one

vividus-plugin-web-app/src/main/java/org/vividus/bdd/steps/ui/web/SetContextSteps.java
```diff
@@ -255,18 +173,18 @@ public void switchingToDefault()
      * </ul>
      * @see <a href="https://en.wikipedia.org/wiki/HTML_element#Frames"><i>Frames</i></a>
      * @see <a href="https://www.w3schools.com/tags/default.asp"><i>HTML Element Reference</i></a>
-     * @param xpathExpression 'xpath' expression to locate frame element
+     * @param locator to locate frame element
```
uarlouski, line 176: If we walk through javadocs of JDK/Spring we can notice that the word after parameter name is always either article or noun, can we apply that common convention in vividus as well?

valfirst: it's not something that can be controlled via tools, that should be mostly internal rules, but it's out of scope for this PR

@uarlouski will you create a ticket and describe the rules? after that all Javadocs could be refactored

ikalinin1: 1. I want to see the profit of the rules;
2. I am pretty sure we could find many differences between spring and JDK conventions;
3. The convention of the steps javadocs should be revised because since we have documentation in place it would be nice to have some standard and clean javadocs to refer them in both directions;
As example:
https://twitter.com/tagir_valeev/status/1259516164985835520

uarlouski: we are talking here only about @param indentifier in java docs, all the mentioned project use @param followed by noun or article word. If just looking at some example from generated javadocs
Parameters:
name - to check for aliases
Parameters:
name - the bean name to check for aliases
Obviously the second case sound more correct

ikalinin1: Per our discussion, fill the ticket and we will discuss Javadocs format changes

PR #593, opened 2020-05-19
[allure-adaptor] Add index to history key for scenarios with same title

### LINKED ISSUE #591: Scenarios with same name are not executed
In case a story has 1+ scenarios with the same scenario name, only one (first one) is executed regardless of whether the steps are identical or not.

STR:
1. Please add 2 scenarios with similar name and different steps.
2. Launch the story

Actual results:
Only first scenario is executed.

Expected results:
Both scenarios should be executed.

valfirst:
Fixes #591


vividus-allure-adaptor/src/main/java/org/vividus/bdd/report/allure/AllureStoryReporter.java
```diff
@@ -558,15 +558,26 @@ private void startTestCase(RunningScenario runningScenario, StoryExecutionStage
         allureRunContext.setScenarioExecutionStage(ScenarioExecutionStage.BEFORE_STEPS);
     }
 
-    private String getHistoryId(String title)
+    private String getHistoryId()
```
ikalinin1, line 561: you could pass RunningScenario as a parameter?

PR #604, opened 2020-05-22
[bdd-engine] Use TableParsers to parse table as string into object representation

uarlouski:


vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/FilteringTableTransformer.java
```diff
@@ -57,18 +52,13 @@ public String transform(String tableAsString, TableParsers tableParsers, TablePr
...
-                .map(LinkedHashMap::new)
-                .map(Map::values)
-                .map(ArrayList::new)
-                .collect(Collectors.toList());
+        List<String> filteredColumns = getFilteredHeaders(byMaxColumns, byColumnNames, tableRows.getHeaders());
+        List<Map<String, String>> result = filterByHeaders(filteredColumns,
+                getFilteredRows(byMaxRows, tableRows.getRows()));
 
-        return ExamplesTableProcessor.buildExamplesTable(filteredColumns, resultRows, properties, true, true);
+        return ExamplesTableProcessor.buildExamplesTable(filteredColumns, result, properties);
```
valfirst, line 61: why not `ExamplesTableProcessor.buildExamplesTable(filteredColumns, result, properties, true, true);` ?

vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/SortingTableTransformer.java
```diff
@@ -34,28 +36,28 @@
...
+                        if (headerValues.contains(column))
+                        {
+                            result = r1.get(column).compareTo(r2.get(column));
+                        }
                     }
                     return result;
                 })
                 .collect(Collectors.toList());
-        return ExamplesTableProcessor.buildExamplesTable(headerValues, rows, properties, true);
+        return ExamplesTableProcessor.buildExamplesTable(headerValues, rows, properties);
```
valfirst, line 61: what is the reason of this change?

uarlouski: As to transformers, we know that TableProperties object contains actual separators and other table properties so there is no need in additional separator checks

vividus-bdd-engine/src/main/java/org/vividus/bdd/transformer/FilteringTableTransformer.java
uarlouski: are you suggesting to refactor all the existing classes that rely on List<List<String>>?

PR #620, opened 2020-05-26
[plugin-web-app] Allow empty arguments for click, double click, click and hold and release actions

uarlouski:


vividus-plugin-web-app/src/main/java/org/vividus/bdd/converter/StringToListSequenceActionConverter.java
```diff
@@ -51,8 +51,13 @@ public StringToListSequenceActionConverter(StringToSearchAttributesConverter str
             .map(params ->
             {
                 SequenceActionType actionType = params.valueAs("type", SequenceActionType.class);
+                String argumentAsString = argumentAs(params, String.class);
+                if (argumentAsString.isEmpty() && actionType.isNullable())
```
ikalinin1, line 55: and what about not nullable arguments? Shouldn't you notify the user about an invalid argument?

PR #636, opened 2020-05-31
[bdd-engine] Allow to set ignore failure per batch

### LINKED ISSUE #583: To add possibility to stop running further batches of tests if the first batch failed e.g: bdd.ignore-failure-in-batch-1=false
Due to the problem that a failed scenario about logging in the application (for UI) or getting SSO token (for API) blocks further pipelines running and even postpones Jenkins restart, we need to have further running of the rest test bathes to be stopped, **if the first batch about login or getting token failed.**
 

We have a property 
**bdd.ignore-failure-in-batches=false**, to customize it for one batch only: e.g 
**bdd.ignore-failure-in-batch-1=false**

uarlouski:
Closes #583

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/BatchedEmbedder.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/BatchedEmbedder.java
@@ -45,12 +45,12 @@ public class BatchedEmbedder extends Embedder
     private final IBddVariableContext bddVariableContext;
     private final BatchStorage batchStorage;
 
-    private boolean ignoreFailureInBatches;
     private boolean reportBeforeStories = true;
     private boolean reportAfterStories;
     private boolean generateViewAfterBatches;
 
     private String batch;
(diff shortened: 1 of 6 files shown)


vividus-bdd-engine/src/main/java/org/vividus/bdd/BatchedEmbedder.java
```diff
@@ -106,7 +108,7 @@ public void runStoriesAsPaths(Map<String, List<String>> storyPathsBatches)
                     storyManager().runStoriesAsPaths(storyPaths, filter, failures);
 
                     handleFailures(failures);
-                    if (!ignoreFailureInBatches && !failures.isEmpty())
+                    if (!batchExecutionConfiguration.isIgnoreFailure() && !failures.isEmpty())
```
valfirst, line 111: `batchExecutionConfiguration.isIgnoreFailure()` -> `ignoreFailure`

PR #668, opened 2020-06-12
[bdd-engine] Fix variables with comma and default value processing

### LINKED ISSUE #667: Variables in lower dot case with default values are not properly defined
Define variable named in lower dot case:
```
When I initialize the SCENARIO variable `a.b` with value `a.b`
```
or
```
bdd.variables.global.a.b=a.b
```
**Actual:**
Obtain _default_ variable value for the defined previously variable:
```
When I initialize the SCENARIO variable `ab` with value `${a.b}`       //a.b
When I initialize the SCENARIO variable `ab` with value `${a.b:NULL}`  //NULL
```
**Expected:**
Obtain variable value for the defined previously variable:
```
When I initialize the SCENARIO variable `ab` with value `${a.b:NULL}`  //a.b
```

ikalinin1:
closes #667


ikalinin1:
Retry DeepCode

PR #685, opened 2020-06-18
[plugin-web-app] Supply meaningful message in case when context eleme…

ikalinin1:
…nt becomes stale

### DIFF
--- a/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/ExpectedSearchContextConditions.java
+++ b/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/ExpectedSearchContextConditions.java
@@ -17,9 +17,11 @@
 package org.vividus.ui.web.action;
 
 import java.util.List;
+import java.util.function.Supplier;
 
 import org.openqa.selenium.By;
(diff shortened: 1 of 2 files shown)


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/ExpectedSearchContextConditions.java
```diff
@@ -17,28 +17,56 @@
...
+    {
+        private static final long serialVersionUID = 2449857900556530700L;
+
+        private StaleContextException(Throwable throwable)
+        {
+            super("Search context used for search is stale. "
+                + "Please double check the tests. "
+                + "You have a few options: "
+                + "1. reset context; "
+                + "2. Synchronize the tests to wait for context's stabilization;", throwable);
```
uarlouski, line 69: would not line breaks make the message readable?

ikalinin1: image
vs
image

Looks better with linebreaks for me

PR #709, opened 2020-06-30
Fix clicking on IE using nativeEvents capability

AlenaBudevich:


vividus-plugin-web-app/src/main/java/org/vividus/selenium/WebDriverType.java
```diff
@@ -81,7 +81,8 @@ public void prepareCapabilities(DesiredCapabilities desiredCapabilities)
         {
             // Workaround for IExplore:
             // https://stackoverflow.com/questions/50287435/actions-movetoelement-not-working-on-ie-11
-            desiredCapabilities.merge(new InternetExplorerOptions().requireWindowFocus());
+            // https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/4403
+            desiredCapabilities.merge(new InternetExplorerOptions().requireWindowFocus().disableNativeEvents());
```
valfirst, line 85: does hovering work with native events disabled?

AlenaBudevich: I think yes. I can share scenario for verifying this.

PR #734, opened 2020-07-13
[bdd-engine] Add verification of steps examples table

uarlouski:


vividus-bdd-engine/src/main/java/org/vividus/bdd/parser/StepExamplesTableParser.java
```diff
@@ -55,4 +64,9 @@ public StepExamplesTableParser(ExtendedConfiguration configuration, InjectableSt
         return configuration.stepCollector().collectScenarioSteps(stepsFactory.createCandidateSteps(),
                 scenario, Map.of(), monitor);
     }
+
+    private static String getErrorMessage(List<String> headers)
+    {
+        return headers.isEmpty() ? "empty table" : Joiner.on(", ").join(headers);
```
valfirst, line 70: `java.lang.String#join(java.lang.CharSequence, java.lang.Iterable<? extends java.lang.CharSequence>)`?

PR #736, opened 2020-07-13
[GithubActions] Setup Github Super-Linter workflow

valfirst:


valfirst:
We use custom fine-tuned linters configured in VIVIDUS build system, Super-Linter is too generic.

PR #780, opened 2020-07-28
[bdd-engine] Add possibility to use java.lang.Math and org.apache.com…

ikalinin1:
…mons.lang3.StringUtils as namespaces

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/EvalExpressionProcessor.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/EvalExpressionProcessor.java
@@ -17,6 +17,7 @@
 package org.vividus.bdd.expression;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Map;
 import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/EvalExpressionProcessor.java
```diff
@@ -35,10 +37,11 @@
     private static final Pattern EVAL_PATTERN = Pattern.compile("^eval\\((.*)\\)$", Pattern.CASE_INSENSITIVE
             | Pattern.DOTALL);
     private static final int EVAL_GROUP = 1;
+    private static final Map<String, Object> NAMESPACES = Map.of("math", Math.class, "string", StringUtils.class);
```
valfirst, line 40: `"string"` ->. `"stringUtils"`?

`string` could be ambiguous

ikalinin1: I was thinking about that too, but it looks long and ugly, maybe we could use something shorter? str_ut ?

valfirst: The practice shows that shortenings don't work well after some time, because it's hard to remember them and sometimes it's not enough clear for new users

PR #782, opened 2020-07-29
[plugin-datetime] Require T designator for time components

uarlouski:


vividus-plugin-datetime/src/test/java/org/vividus/bdd/expression/DateExpressionProcessorTests.java
```diff
@@ -103,6 +103,7 @@ void shouldAccept(String expression)
             "generateDate(PT1)",
             "generateDate(P10M12)",
             "generateDate(P10)",
+            "generateDate(P10S)",
```
ikalinin1, line 106: Add/Update integration tests?

uarlouski: There is already test Expressions -> Verify date generation and format that checks the expression works from user side, more specific combinations of date and time symbols I would prefer to test using unit tests

PR #801, opened 2020-08-05
Add step adding headers to proxied requests with given URL pattern

### LINKED ISSUE #603: [P3] Need to impose header to exclude test runs from analytics
Currently, new approach of exclusion of testing activities from analytics is to be implemented. We need to add new header not only to API tests, but also to UI test.

Proposed solution is https://stackoverflow.com/questions/15645093/setting-request-headers-in-selenium

ngrudnitsky:
Closes #603


vividus-tests/src/main/resources/story/integration/ProxyStepsTests.story
```diff
@@ -31,3 +31,10 @@ Scenario: Verify step When I wait until HTTP $httpMethod request with URL patter
 Given I am on a page with the URL 'http://httpbin.org/get'
 When I wait until HTTP GET request with URL pattern `http://httpbin.org/get` exists in proxy log
 Then number of HTTP GET requests with URL pattern `http://httpbin.org/get` is EQUAL TO `1`
+
+Scenario: Verify step When I add headers:$headers to proxied requests with URL pattern $urlPattern
+When I add headers to proxied requests with URL pattern `http://httpbin.org/headers`:
+|name    |value    |
+|testName|testValue|
+Given I am on a page with the URL 'http://httpbin.org/headers'
+When I save response body to the SCENARIO variable 'requestHeaders'
```
ikalinin1, line 40: Verification is missing?

vividus-plugin-web-app/src/main/java/org/vividus/proxy/steps/ProxySteps.java
```diff
@@ -238,6 +241,30 @@ public String toString()
...
     }
 
+    /**
+     * Add headers to proxy request that will be used while sending request with given urlPattern
+     * @param headers ExamplesTable representing list of headers with columns "name" and "value" specifying HTTP header
+     * names and values respectively
+     * @param urlPattern The string value of URL-pattern to filter by
+     */
+    @When("I add headers to proxied requests with URL pattern `$urlPattern`:$headers")
+    public void addHeadersToProxyRequest(ExamplesTable headers, String urlPattern)
```
ikalinin1, line 251: `String urlPattern` why not use some flexible mechanism to match URLs like Patter or StringComparisonRule ?

ikalinin1:
@ngrudnitsky  feel free to contact me in skype ivan_kalinin_by if you still in need of help

vividus-plugin-web-app/src/main/java/org/vividus/proxy/steps/ProxySteps.java
```diff
@@ -238,6 +243,32 @@ public String toString()
...
+    public void addHeadersToProxyRequest(StringComparisonRule comparisonRule, String urlPattern, ExamplesTable headers)
+    {
+        DefaultHttpHeaders httpHeaders = new DefaultHttpHeaders();
+        headers.getRowsAsParameters(true)
+                .forEach(row -> httpHeaders.add(
+                        row.valueAs("name", String.class),
+                        (Object) row.valueAs("value", String.class)
+                ));
+        Matcher<String> expected = comparisonRule.createMatcher(urlPattern);
+        proxy.addRequestFilter((request, contents, messageInfo) -> {
```
ikalinin1, line 263: @valfirst  in the current implementation, the filter will be cleared off on before scenario hook.
Is it expected behavior?
Shouldn't we support filters that will work whole run time? (For example, if we need to pass basic authentication for some of the requests)
And I think we should give the user the possibility to clear the filters via step.

valfirst (approved):

vividus-plugin-web-app/src/main/java/org/vividus/proxy/steps/ProxySteps.java
valfirst: that's a good point, but it's out of scope this PR

vividus-plugin-web-app/src/test/java/org/vividus/proxy/steps/ProxyStepsTests.java
```diff
@@ -261,6 +267,17 @@ void testWaitRequestInProxyLog()
...
 
+    @Test
+    void testAddHeadersToProxyRequest()
+    {
+        HttpRequest request = mock(HttpRequest.class);
+        HttpMessageContents contents = mock(HttpMessageContents.class);
+        HttpMessageInfo message = mock(HttpMessageInfo.class);
+        ExamplesTable headers = new ExamplesTable("|name|value|\n|name1|value1|");
+        proxySteps.addHeadersToProxyRequest(StringComparisonRule.IS_EQUAL_TO, URL_PATTERN, headers);
+        verify(proxy).addRequestFilter(argThat(filter -> filter.filterRequest(request, contents, message) == null));
```
valfirst, line 278: I believe unit test should be expanded a bit to invoke the filter actually (and 2 unit tests will be required: when URL matches and doesn't match to the rule)

vividus-tests/src/main/resources/story/integration/ProxyStepsTests.story
```diff
@@ -31,3 +31,18 @@ Scenario: Verify step When I wait until HTTP $httpMethod request with URL patter
 Given I am on a page with the URL 'http://httpbin.org/get'
 When I wait until HTTP GET request with URL pattern `http://httpbin.org/get` exists in proxy log
 Then number of HTTP GET requests with URL pattern `http://httpbin.org/get` is EQUAL TO `1`
+
+Scenario: Verify step When I add headers to proxied requests with URL pattern which $comparisonRule `$urlPattern`:$headers
```
ikalinin1, line 35: Please add requirenentId

PR #803, opened 2020-08-06
[soft-assert] Add possibility to disable potentially known issues

### LINKED ISSUE #800: Add a possibility to enable/disable potentially known issue tracking in Allure
When known issue functionality is used in test suite, some of the failures that partially meet known issues criteria a marked as potentially known issues.

In the cases of API scenarios, where a lot of assertions in known issues are for status codes, similar status codes are marked as potentially known issues which is misleading.

Please add a possibility to disable/enable potentially known issue tracking mechanism

ikalinin1:
Closes #800

### DIFF
--- a/vividus-soft-assert/src/main/java/org/vividus/softassert/issue/KnownIssueChecker.java
+++ b/vividus-soft-assert/src/main/java/org/vividus/softassert/issue/KnownIssueChecker.java
@@ -31,6 +31,7 @@ public class KnownIssueChecker implements IKnownIssueChecker
 
     private ITestInfoProvider testInfoProvider;
     private IIssueStateProvider issueStateProvider;
+    private boolean detectPotentiallyKnown;
 
     public KnownIssueChecker(IKnownIssueProvider knownIssueProvider, KnownIssueDataProvider knownIssueDataProvider)
     {
@@ -79,6 +80,11 @@ public class KnownIssueChecker implements IKnownIssueChecker
         this.issueStateProvider = issueStateProvider;
     }
 
+    public void setDetectPotentiallyKnown(boolean detectPotentiallyKnown)
+    {
+        this.detectPotentiallyKnown = detectPotentiallyKnown;
+    }
+
     private class CandidateIssue
     {
         private static final int ALL_PATTERNS = 3;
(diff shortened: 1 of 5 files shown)


vividus-soft-assert/src/main/java/org/vividus/softassert/issue/KnownIssueChecker.java
```diff
@@ -31,6 +31,7 @@
 
     private ITestInfoProvider testInfoProvider;
     private IIssueStateProvider issueStateProvider;
+    private boolean allowPotentiallyKnownIssues;
```
uarlouski, line 34: hmm imho enable/disable looks semantically better here

ikalinin1: We'll wait @valfirst  on this comment :)

valfirst: `detectPotentiallyKnownIssues`?

vividus/src/main/resources/properties/defaults/default.properties
```diff
@@ -76,6 +76,7 @@ system.allure.results.directory=${output.directory}/allure-results
 start-context-listener.cleanable-directories=${output.report-directory}
 
 known-issue-provider.fileName=known-issues.json
+known-issue-checker.allow-potentially-known-issues=true
```
valfirst, line 79: for users `known-issue-provider` or `known-issue-checker` don't have any difference, I propose to keep naming like `known-issue-provider...` (even if it belongs to another bean internally)

vividus-soft-assert/src/main/java/org/vividus/softassert/issue/KnownIssueChecker.java
valfirst: also would you mind updating/adding docs?

ikalinin1: @valfirst it could be done as separate commit to unblock our guys, or we can wait until the docs are ready?

valfirst: I think they will wait for release, as for docs: you need to add only description or the new property, nothing else

PR #806, opened 2020-08-07
Rename known issues file name property

ikalinin1:
I'll use the same prefix for the new property in #803


vividus/src/main/resources/org/vividus/spring-vividus.xml
```diff
@@ -18,7 +18,7 @@
     </bean>
 
     <bean class="org.vividus.bdd.issue.KnownIssueProvider" init-method="init">
-        <property name="fileName" value="${known-issue-provider.fileName}" />
+        <property name="fileName" value="${known-issues.file-name}" />
```
valfirst, line 21: I would prefer to use the old naming, since `file-name` is not an attribute of `known-issues`

PR #814, opened 2020-08-12
Fix handling of variables of different nesting step-levels

### LINKED ISSUE #763: iterationVariable isn't extracted after condition step
When we use cycle step with condition inside $iterationVariable isn't extracted after this condition.

**Steps to reproduce:**
```gherkin
When I execute steps while counter is <= `1` with increment `1` starting from `1`:
{headerSeparator=!,valueSeparator=!}
!step!
!When I initialize the SCENARIO variable `test1` with value `${iterationVariable}`!
!When the condition 'true' is true I do!
!|step|!
!|When I initialize the SCENARIO variable `value` with value `test`|!
!When I initialize the SCENARIO variable `test2` with value `${iterationVariable}`!
!Then `${test1}` is = `${test2}`!
```

**Expected behavior**
Variables ${test1} = ${test2}

**Artifacts**
Allure report: [iterationVariable.zip](https://github.com/vividus-framework/vividus/files/4964407/iterationVariable.zip)

**Execution environment:**
Vividus version: 0.2.5

### LINKED ISSUE #805: Value is not retrieved from table when composite step is put inside of 'When the condition' step
Steps to reproduce:
1. Create table with variable in it 
e.g.
```
{transformer=FROM_LANDSCAPE}
|number      |100   |
```
2. Create composite step which includes steps **with reference to table variables**
e.g.
```
Composite: When I initialize variable for number
When I initialize the story variable `someVariable` with value `<number>`
```
3. Create story where composite step is put into  'When the condition' step, e.g.
```
Lifecycle:
Examples:
/regression/story/variables.table

Scenario: Condition
When the condition '#{eval(`1` == `1`)}' is true I do
|step|
|When I initialize variable for number|
When I initialize variable for number
```
4. Run it and observe results:
(see att. composite_in_condition)
```
When the condition '#{eval(`1` == `1`)}' is true I do |step| |When I initialize variable for number|
-> When I initialize variable for number
     -> When I initialize the ｟story｠ variable `someVariable` with value `<number>`
              ->Saving a value '<number>' into the 'STORY' variable 'someVariable' 
```

**Actual result**:
- value **100** from table is not retrieved from table when composite step **is put into**  'When the condition' step
- value **100** from table is get successfully from table **when there is no**  'When the condition' step

**Expected result**: Value should be retrieved successfully from table **when composite step is put inside** of 'When the condition' step

composite_in_condition
[composite.steps.txt](https://github.com/vividus-framework/vividus/files/5036065/composite.steps.txt)
[Composite-in-condition-allure.zip](https://github.com/vividus-framework/vividus/files/5036066/Composite-in-condition-allure.zip)
[logs.txt](https://github.com/vividus-framework/vividus/files/5036067/logs.txt)
[Test.story.txt](https://github.com/vividus-framework/vividus/files/5036068/Test.story.txt)
[variables.table.txt](https://github.com/vividus-framework/vividus/files/5036069/variables.table.txt)

valfirst:
Fixes #763 

Since Vividus supports nested steps, the variables of step-level should be initialized and cleaned up according to their level.
Previously the variables of different step levels were stored in a single collection, which was cleaned up after each step, as result the nested steps could remove variables created at wrapping steps.


vividus-bdd-engine/src/main/java/org/vividus/bdd/variable/Variables.java
```diff
@@ -16,27 +16,170 @@
...
+    private static final int LIST_INDEX_GROUP = 2;
+    private static final int MAP_KEY_GROUP = 3;
+    private static final Pattern COMPOUND_VARIABLE_PATTERN = Pattern.compile(
+            "([^\\[\\].:]+):?(?:\\[(\\d+)])?:?(?:\\.([^:]+))?:?");
 
-    public Variables()
+    private final Map<String, Object> batchVariables;
+    private final Map<String, Object> storyVariables;
+    private final Map<String, Object> scenarioVariables;
+    private final Deque<Map<String, Object>> stepVariables;
```
uarlouski, line 41: have you considered using arraydeque?

vividus-bdd-engine/src/main/java/org/vividus/bdd/variable/Variables.java
```diff
@@ -16,27 +16,170 @@
...
+
+    public Variables(Map<String, Object> batchVariables)
+    {
+        this.batchVariables = batchVariables;
+        storyVariables = new HashMap<>();
+        scenarioVariables = new HashMap<>();
+        stepVariables = new LinkedList<>();
```
uarlouski, line 48: have you considered using arraydeque?

valfirst: need to be benchmarked

PR #834, opened 2020-08-19
Verify elements in state existence

ikalinin1:


AlenaBudevich (approved):

vividus-plugin-web-app/src/main/java/org/vividus/bdd/steps/ui/web/ElementSteps.java
```diff
@@ -315,6 +317,25 @@ public void clickOnAllElements(SearchAttributes locator)
...
+     * EQUAL_TO</b></i>)
+     * @param quantity Desired amount of elements
+     */
+    @Then("number of $state elements found by `$locator` is $comparisonRule `$quantity`")
+    public void doesElementsNumberInStateExists(State state, SearchAttributes locator, ComparisonRule comparisonRule,
+            int quantity)
+    {
+        String description = "Element is " + state;
+        thePageContainsQuantityElements(locator, comparisonRule, quantity)
+                .forEach(e -> baseValidations.assertElementState(description, state, e));
```
valfirst, line 336: Use case:
```
Then number of visible elements found by `...` is equal to `3`
```

There are 4 elements by locator, and only 3 of them are visible, what will the step result be? I assume failure

ikalinin1: Yep as stated - equal to 3 visible elements expected but got 4: 3 visible, and 1 invisible.
1. Fail condition check
2. Fail visibility check

PR #860, opened 2020-08-27
[tests] Cover drop down steps with integration tests

ikalinin1:


vividus-tests/src/main/resources/story/integration/DropdownStepsTests.story
```diff
@@ -0,0 +1,51 @@
+Description: Integration tests for DropdownSteps class.
+
+Meta:
+    @epic vividus-plugin-web-app
+
+Scenario: Set-Up
```
valfirst, line 6:
```
Lifecycle:
Before:
Scope: STORY
```
?

valfirst (changes requested):

vividus-tests/src/main/resources/story/integration/DropdownStepsTests.story
valfirst: NPE, it'll be fixed later

valfirst (approved):

uarlouski (approved):

PR #862, opened 2020-08-27
Fix NPE when browser session starting in before story

ikalinin1:


vividus-plugin-web-app/src/main/java/org/vividus/bdd/monitor/ScreenshotOnFailureMonitor.java
```diff
@@ -123,8 +126,10 @@ private boolean isStoryHasNoScreenshotOnFailureMeta()
 
     private boolean isScenarioHasNoScreenshotsOnFailureMeta()
     {
-        RunningStory runningStory = bddRunContext.getRunningStory();
-        return runningStory.getRunningScenario().getScenario().getMeta()
-                .hasProperty(NO_SCREENSHOT_ON_FAILURE_META_NAME);
+        return Optional.ofNullable(bddRunContext.getRunningStory())
```
valfirst, line 129: when can `runningStory` be `null`?

vividus-plugin-web-app/src/main/java/org/vividus/bdd/monitor/ScreenshotOnFailureMonitor.java
```diff
@@ -123,8 +126,10 @@ private boolean isStoryHasNoScreenshotOnFailureMeta()
...
+                       .map(RunningStory::getRunningScenario)
+                       .map(RunningScenario::getScenario)
```
valfirst, line 131: similar here: if `runningScenario` is not `null`, will `scenario` ever be `null`?

valfirst (changes requested):

valfirst (approved):

uarlouski (approved):

PR #902, opened 2020-09-07
Break circular reference of Spring beans

uarlouski:


vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/ElementActionService.java
```diff
@@ -20,16 +20,18 @@
...
+    @Inject private Set<IElementAction> elementActions;
 
-    private final Set<LocatorType> searchLocatorTypes;
-    private final Set<LocatorType> filterLocatorTypes;
+    private Set<LocatorType> searchLocatorTypes;
+    private Set<LocatorType> filterLocatorTypes;
 
-    public ElementActionService(Set<IElementAction> elementActions)
+    @SuppressWarnings("unused")
+    private void init()
```
valfirst, line 33: I recommend to keep it `public` since `private` makes this class unusable outside Spring

valfirst (changes requested):

valfirst (approved):

PR #910, opened 2020-09-08
Update plugin-excel.adoc

AlenaSikorskaya:
please see the changes

### DIFF
--- a/docs/modules/plugins/pages/plugin-excel.adoc
+++ b/docs/modules/plugins/pages/plugin-excel.adoc
@@ -68,3 +68,8 @@ Examples:
 Examples:
 {transformer=FROM_EXCEL, path=/my-excel-file.xlsx, sheet=Data, addresses=A2;B4;A7, column=test, joinValues=true}
 ----
+
+=== ALLURE REPORT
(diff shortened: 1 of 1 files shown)


docs/modules/plugins/pages/plugin-excel.adoc
```diff
@@ -68,3 +68,8 @@ Examples:
 Examples:
 {transformer=FROM_EXCEL, path=/my-excel-file.xlsx, sheet=Data, addresses=A2;B4;A7, column=test, joinValues=true}
 ----
+
+=== ALLURE REPORT
```
ikalinin1, line 72: It's not clear what functionality is referenced

AlenaSikorskaya: this is Allure func, if you want to see certain values near by testcase (and not to find this value on the right). sorry, I can't formulate better)

valfirst: there is no ability to set the name for test based on the ExamplesTable iteration, the mentioned trick can be used as workaround for now, but we can't put it to the official documentation

PR #914, opened 2020-09-09
Load converters depending on profile

uarlouski:


AlenaBudevich (approved):

vividus-plugin-web-app/src/test/java/org/vividus/bdd/converter/StringToListSequenceActionConverterTests.java
```diff
@@ -33,6 +33,8 @@
 import org.mockito.Mock;
```
valfirst, line 33: please update the package of this class

valfirst (changes requested):

valfirst (approved):

PR #916, opened 2020-09-09
[plugin-mobile-app] Use mobile_app spring profile only if it's selected by vividus mobile related profile

uarlouski:


AlenaBudevich (approved):

vividus-plugin-mobile-app/src/main/resources/properties/defaults/run.properties
```diff
@@ -1,5 +1,3 @@
-spring.profiles.active=mobile_app
```
ikalinin1, line 1: what is the reason for these changes?

vividus-plugin-mobile-app/src/main/resources/properties/defaults/run.properties
```diff
@@ -1,5 +1,3 @@
...
-
 bdd.all-meta-filters=groovy: !skip && (${bdd.meta-filters})
 
 mobile-app=true
```
ikalinin1, line 3: What about this one?

ikalinin1 (changes requested):

vividus-plugin-mobile-app/src/main/resources/properties/defaults/run.properties
uarlouski: mobile_app profile is used by default when we add plugin-mobile-app as a dependency, so it should not land in defaults folder

valfirst (approved):

ikalinin1 (approved):

PR #917, opened 2020-09-09
[vividus] Fix web integration tests to use web spring profile

uarlouski:


uarlouski:
Closed for investigations

PR #918, opened 2020-09-09
[vividus] Add integration tests for mobile profiles

uarlouski:


vividus/src/integrationTest/resources/properties/environment/integrationtest/run.properties
```diff
@@ -1 +1,3 @@
 bdd.story-loader.batch-1.resource-location=story
+
+selenium.grid.url=https://seleniumgrid.io/wd/hub
```
valfirst, line 3: it's ok for this PR, it will require rework in future

valfirst, line 3: shouldn't it be put under mobile_app profile for this integration tests?

valfirst (approved):

PR #921, opened 2020-09-10
fix for vividus/issues/883

### LINKED ISSUE #883: I want "round" function might support numbers like "1.056E7b" or Vividus can correctly work with "big" numbers like 10 560 000 and alike without adding 'b'
There is a problem with numbers which have more than 5 digits (and which might look like "1.056E7b" after vividus` interpreting)
At the moment we use such with adding 'b' after it. 

Example: 
```
${x} * ${y} = ${x*y} in order to use variable ${x*y} in further calculations we add 'b' => '${x*y}b' 
132000*80 = 10 560 000 == 1.056E7
```

It would be great if mentioned numbers type will correctly collaborate with "round" functionality. 

Also I would like to notice that number in the variable '${x*y}b' (10 560 000 == 1.056E7) is sent with wrong value into field in used UI. 
Example: 
1) save expected value in '${x*y}'
2) using step 'When I enter `${x*y}b` in field located `By.id(bv1)' 
3) send request with value
it does not sent 1.056E7 but delete everything after letter (looks like 1.056) and automattically round number => 1.06
but in the same time, value of this variable is used in the possible next steps with its value = 1.056E7

BViktoryia:
Fixes #883

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
@@ -38,8 +38,8 @@ public class RoundExpressionProcessor implements IExpressionProcessor
     {
         String roundingModes = Stream.of(RoundingMode.values()).map(e -> e.name().toLowerCase())
                 .collect(Collectors.joining("|"));
-        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?[Ee]?-?\\+?\\d*)(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$",
-            roundingModes);
+        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?[Ee]?-?\\+?\\d*)"
+                        + "(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$", roundingModes);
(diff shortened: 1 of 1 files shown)


BViktoryia:
fix for vividus/issues/883

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
```diff
@@ -38,8 +38,8 @@
     {
         String roundingModes = Stream.of(RoundingMode.values()).map(e -> e.name().toLowerCase())
                 .collect(Collectors.joining("|"));
-        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?)(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$",
-            roundingModes);
+        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?[Ee]?-?\\+?\\d*)"
+                        + "(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$", roundingModes);
```
uarlouski, line 42: please add unit and integration tests

uarlouski (commented):
Commit name should reflect made changes without opening linked issue

valfirst:
Superseded by #1018

PR #930, opened 2020-09-11
[extension-selenium] Notify user about null search context usage

ikalinin1:


uarlouski (approved):

vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/AbstractElementAction.java
```diff
@@ -59,6 +59,7 @@ public LocatorType getType()
         {
             return findElements(searchContext, locator, parameters, false);
         }
+        LOGGER.error("Unable to locate elements using `null` search context");
```
valfirst, line 62:
```
"Unable to locate elements using `null` search context"
```
 -> 
`Unable to locate elements, because search context is not set` 
?

valfirst (changes requested):

valfirst (approved):

PR #932, opened 2020-09-11
[vividus] Extract common logic to new BeanFactoryUtils

valfirst:


vividus/src/main/java/org/vividus/spring/BeanFactoryUtils.java
```diff
@@ -0,0 +1,60 @@
...
+                .flatMap(List::stream)
+                .collect(toList());
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <K, V> Map<K, V> mergeMaps(ListableBeanFactory beanFactory, String mapBeanNamePrefix)
+    {
+        return streamOfBeansByNamePrefix(beanFactory, mapBeanNamePrefix, Map.class)
+                .map(m -> (Map<K, V>) m)
+                .flatMap(m -> m.entrySet().stream())
```
uarlouski, line 49: what about Map::entrySet and then flatMap Set::stream (just from point of view of stream-diseased guy ;) )

uarlouski (approved):

PR #938, opened 2020-09-14
[extension-selenium] Validate search context before search

ikalinin1:


vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/AbstractElementAction.java
```diff
@@ -33,6 +33,8 @@
 
 public abstract class AbstractElementAction implements IElementAction
 {
+    public static final String NOT_SET_CONTEXT = "Unable to locate elements, because search context is not set";
```
valfirst, line 36: it's better to move it to interface

valfirst (changes requested):

valfirst (approved):

uarlouski (approved):

PR #948, opened 2020-09-16
Fix locators according to updates in react native app

uarlouski:


AlenaBudevich (approved):

valfirst:
The actual fix is https://github.com/vividus-framework/vividus-test-app/pull/30

PR #978, opened 2020-10-02
Disallow possibility to check then number of elements in state when l…

### LINKED ISSUE #886: Improve debug information for step Then number of state elements found
The problem is a possible conflict between checking the visibility state and the provided locator.
For example:
Then number of NOT_VISIBLE elements found by `By.xpath(//button[text()="Sometext"])` is EQUAL_TO `1`
In this case we're trying to check invisibility of elements among the elements filtered by the default visibility state: visible.

Add validation for this kind of collision and provide some additional information into logs\report. (Maybe it'll be better to spawn an additional error message)

ikalinin1:
…ocator configured for the same state

closes #886


valfirst (changes requested):
The failed integration test should be fixed

valfirst (approved):

uarlouski (approved):

PR #1018, opened 2020-10-13
Added support for numbers with mantissa

### LINKED ISSUE #883: I want "round" function might support numbers like "1.056E7b" or Vividus can correctly work with "big" numbers like 10 560 000 and alike without adding 'b'

BViktoryia:
Closes #883


ikalinin1 (approved):

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
```diff
@@ -38,8 +38,8 @@
     {
         String roundingModes = Stream.of(RoundingMode.values()).map(e -> e.name().toLowerCase())
                 .collect(Collectors.joining("|"));
-        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?)(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$",
-            roundingModes);
+        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?[Ee]?-?\\+?\\d*)"
```
valfirst, line 41: this regex will also match the following invalid strings:
 - `0.15237E`
 - `0.15237E-+2`
 - `0.15237-2`
 - `0.15237+2`

valfirst (changes requested):

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
BViktoryia: fixed

valfirst: I believe all items from the comment above should be added to negative unit tests

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
```diff
@@ -38,8 +38,8 @@
     {
         String roundingModes = Stream.of(RoundingMode.values()).map(e -> e.name().toLowerCase())
                 .collect(Collectors.joining("|"));
-        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?)(?:,\\s*(\\d+))?(?:,\\s*(%s))?\\))$",
-            roundingModes);
+        String pattern = String.format("^round(?:\\((-?\\d+(?:\\.\\d*)?(?:[Ee]+-?\\d+)?)(?:,\\s*(\\d+))?"
```
valfirst, line 41: the pattern won't match `0.15237E+2` which is a valid number

valfirst (changes requested):

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
BViktoryia: added

vividus-bdd-engine/src/main/java/org/vividus/bdd/expression/RoundExpressionProcessor.java
BViktoryia: fixed

valfirst (approved):

PR #1019, opened 2020-10-13
[plugin-mobile-app] Add step to clear field

uarlouski:


docs/modules/plugins/pages/plugin-mobile-app.adoc
```diff
@@ -321,6 +321,36 @@ Scenario: Type text
 When I type `Bob` in field located `accessibilityId(username)`
 ----
 
+=== Clear text
```
valfirst, line 324: Clear field ?

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/action/KeyboardActions.java
```diff
@@ -75,4 +76,23 @@ public void typeText(WebElement element, String text)
...
+     * <ol>
+     * <li>check if the element's text is empty</li>
+     * <li>if the element's text is not empty then clear the element, otherwise do nothing</li>
+     * </ol>
+     * @param element element to clear, must not be {@code null}
+     */
+    public void clearText(WebElement element)
+    {
+        String elementText = element.getText();
+        if (!StringUtils.isEmpty(elementText))
```
valfirst, line 93: why is this check needed?

uarlouski: getText is ~5 times faster than clear command, so if field is empty we will not spend time on clearing it

valfirst (changes requested):

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/action/KeyboardActions.java
uarlouski: Android platform: ~300ms vs ~1600ms

ikalinin1 (approved):

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/action/KeyboardActions.java
ikalinin1: So in the most common case it is 1600 + 300 = 1800 ? :)

valfirst (approved):

PR #1025, opened 2020-10-14
Fix NPE during property to locator conversion

ikalinin1:


vividus-plugin-web-app/src/main/java/org/vividus/spring/StringToLocatorConverter.java
```diff
@@ -22,16 +22,16 @@
 
 public class StringToLocatorConverter implements Converter<String, Locator>
 {
-    private LocatorConversionUtils conversionUtils;
+    private final LocatorConversionUtils conversionUtils;
```
uarlouski, line 25: I might be wrong but this may lead to circular dependencies

valfirst (approved):

vividus-plugin-web-app/src/main/java/org/vividus/spring/StringToLocatorConverter.java
ikalinin1: Sure thing it could, but the build looks ok? Won't usage of setter hide circular dependency if we have any?

uarlouski: - previously it also looked good before that failed first time, I mean that this thing does not reproduce stably
- didn't get your second question :)

uarlouski (approved):

vividus-plugin-web-app/src/main/java/org/vividus/spring/StringToLocatorConverter.java
ikalinin1: Let me rephrase :)
We should not workaround circular deps in most of the cases, they should be broken. So I would like to be aware if we got one here :)

PR #1050, opened 2020-10-21
[tests] Use Vividus test site for ResourceCheckSteps tests

valfirst:
The URL from https://www.example.com/ may behave differently, so we need to use something more stable: Vividus test site

fail

### DIFF
--- a/vividus-tests/src/main/resources/story/integration/ResourceCheckSteps.story
+++ b/vividus-tests/src/main/resources/story/integration/ResourceCheckSteps.story
@@ -6,7 +6,7 @@ Meta:
 Lifecycle:
 Examples:
 |pageToValidate          |
-|https://www.example.com/|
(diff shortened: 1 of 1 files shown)


vividus-tests/src/main/resources/story/integration/ResourceCheckSteps.story
```diff
@@ -5,8 +5,8 @@ Meta:
 
 Lifecycle:
 Examples:
-|pageToValidate          |
-|https://www.example.com/|
+|pageToValidate                                     |
+|http://vividus-test-site.herokuapp.com/windows.html|
```
ikalinin1, line 9: we have a global variable ${vividus-test-site-url}

ikalinin1: @valfirst  ^^

PR #1070, opened 2020-10-27
[plugin-mobile-app] Add text part filter

uarlouski:


docs/modules/plugins/partials/locator-header.adoc
```diff
@@ -35,4 +35,8 @@ By. prefix is optional.
 |Get an element by the specified index from a collection of elements found by a locator. Indexes start from 1.
 |tagName(div)->filter.index(2)
 
+|textPart
+|Get an element by it's text part.
```
valfirst, line 39: `Get` -> `Filter`

vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/IElementFilterAction.java
```diff
@@ -17,10 +17,28 @@
...
 import org.openqa.selenium.WebElement;
 
 public interface IElementFilterAction extends IElementAction
 {
-    List<WebElement> filter(List<WebElement> elements, String value);
+    default UnsupportedOperationException unsupportedMatchesOperationException()
+    {
+        return new UnsupportedOperationException(
+            "The matches(WebElement, String) operation is not supported for the " + getType().getAttributeName()
+                + " filter");
```
valfirst, line 31: I don't think it should be in the interface, since it's the actual implementation
I propose to move it to the class level and even exception message can be skipped

valfirst (changes requested):

vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/GenericTextPartFilter.java
```diff
@@ -14,25 +14,31 @@
...
+    {
+        this.locatorType = locatorType;
+        this.elementActions = elementActions;
+    }
+
+    @Override
+    public boolean matches(WebElement element, String textPart)
     {
-        super(WebLocatorType.TEXT_PART);
+        return elementActions.getElementText(element).contains(textPart);
```
ikalinin1, line 36: Do we really need a "contains", could we just use regex for this?

ikalinin1 (approved):

vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/GenericTextPartFilter.java
uarlouski: I think it can break use cases if someone used in his text part special symbols

valfirst (approved):

PR #1075, opened 2020-10-28
[vividus] Always compare numeric values as numbers

valfirst:


ikalinin1 (approved):

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
```diff
@@ -169,7 +158,15 @@ else if (instanceOfMap(variable1) && instanceOfMap(variable2))
                 return compareListsOfMaps(List.of(variable1), List.of(variable2));
             }
         }
-        return compare(variable1.toString(), condition, variable2.toString());
+        String variable1AsString = variable1.toString();
```
ikalinin1, line 161: I believe Sting.valueOf will be safer here?

valfirst: in general - agree
as for this case: do you have idea when `variable1`/`variable2` can be `null`?

uarlouski (approved):

PR #1110, opened 2020-11-09
[plugin-rest-api] Add ability to create multipart request with dynamic content

### LINKED ISSUE #571: Update step 'Given multipart request' to add ability to create multipart request with dynamic files
As a **tester**
**Given** I need to upload file to resource by PUT/POST request 
**When** I save dynamic content, e.g. table, in $variable
**Then** I want to set multipart request entity using data from $variable 

**Description**:
During test run it is needed to create multiple tables with different content and then upload created file to resource using PUT/POST request.
It is needed to add ability to put required file content directly in multipart request to exclude interaction with file system, to not to depend on way of running tests (eclipse, intellij idea, etc.) 
Current implementation of 'Given multipart request' step doesn't allow me to add content  in request and step only allows to add link to existing file

_Possible implementation_:
Add new columns **filename** and **content**. 
**filename** - random file name with file extension
**content** - variable which includes expected content (e.g. text or table) OR directly desired content

```
When I initialize the story variable `content` with value `a,b,c
1,2,3`
Given multipart request:
|name|type|filename          |content|contentType             |
|file|file| someFileTitle.csv| ${content} |application/vnd.ms-excel|
```
AND
As tables are expected to be send also, the ability to set multi-line files must be considered.
```
Given multipart request:
|name|type|filename        |content      |contentType             |
|fil |file| someFileTitle.csv| a,b,c\n1,2,3 |application/vnd.ms-excel|
```

**Note**: Old implementation of 'Given multipart request' also should continue to work

valfirst:
Closes #571

### DIFF
--- a/vividus-plugin-rest-api/src/main/java/org/vividus/bdd/model/RequestPartType.java
+++ b/vividus-plugin-rest-api/src/main/java/org/vividus/bdd/model/RequestPartType.java
@@ -16,11 +16,13 @@
 
 package org.vividus.bdd.model;
 
-import java.util.Optional;
+import java.nio.charset.StandardCharsets;
 
 import org.apache.commons.io.FilenameUtils;
 import org.apache.http.entity.ContentType;
-import org.apache.http.entity.mime.MultipartEntityBuilder;
+import org.apache.http.entity.mime.content.ByteArrayBody;
+import org.apache.http.entity.mime.content.ContentBody;
+import org.apache.http.entity.mime.content.StringBody;
 import org.vividus.util.ResourceUtils;
 
 public enum RequestPartType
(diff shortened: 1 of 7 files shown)


vividus-plugin-rest-api/src/main/java/org/vividus/bdd/model/RequestPartType.java
```diff
@@ -16,37 +16,87 @@
...
+
+        @Override
+        public ContentBody createPart(String value, ContentType contentType)
         {
             byte[] byteArray = ResourceUtils.loadResourceAsByteArray(getClass(), value);
             String fileName = FilenameUtils.getName(value);
-            builder.addBinaryBody(name, byteArray,
-                    contentType.map(ContentType::parse).orElse(ContentType.DEFAULT_BINARY),
-                    fileName);
+            return new ByteArrayBody(byteArray, contentType, fileName);
```
uarlouski, line 64: createPart(value, contentType, FilenameUtils.getName(value)) maybe

ikalinin1 (approved):

uarlouski (approved):

PR #1156, opened 2020-11-20
[plugin-aws-lambda] Add step to invoke AWS Lambda function

### LINKED ISSUE #1132: Add possibility to invoke lambda
Could you please add possibility to invoke lambda and view results of processing (Log output)

valfirst:
Closes #1132


docs/modules/plugins/pages/plugin-aws-lambda.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+== Steps
+
+=== Invoke a Lambda function
+
+Invoke a Lambda function by providing the function name and the payload to pass to the Lambda function.
+
+[source,gherkin]
+----
+When I invoke Lambda function `$functionName` with payload `$payload` and save result to $scopes variable `$variableName`
```
ikalinin1, line 23: Maybe `Lambda` -> `AWS λ`

vividus-plugin-aws-lambda/src/main/java/org/vividus/aws/lambda/steps/LambdaSteps.java
```diff
@@ -0,0 +1,104 @@
...
+        InvokeResult invokeResult = awsLambdaClient.invoke(invokeRequest);
+
+        Map<String, String> result = new HashMap<>();
+        result.put("payload", new String(invokeResult.getPayload().array(), StandardCharsets.UTF_8));
+        result.put("status-code", invokeResult.getStatusCode().toString());
+        result.put("log-result",
+                new String(Base64.getDecoder().decode(invokeResult.getLogResult()), StandardCharsets.UTF_8));
+        result.put("executed-version", invokeResult.getExecutedVersion());
+        String functionError = invokeResult.getFunctionError();
+        if (functionError != null)
```
ikalinin1, line 98: considering the case when the user will expect an empty function error, wouldn't it be easier to verify an empty string instead of a missing variable?

ikalinin1: Analog of the SSH steps' empty error output stream

ikalinin1 (approved):

docs/modules/plugins/pages/plugin-aws-lambda.adoc
valfirst: `λ` will be painful for users

valfirst: I would even say, not everyone knows what is the name of this letter

vividus-plugin-aws-lambda/src/main/java/org/vividus/aws/lambda/steps/LambdaSteps.java
valfirst: - I've tested this functionality, `functionError` is some error type (https://docs.aws.amazon.com/lambda/latest/dg/invocation-retries.html), I have not found any enumeration of possible values here, I always see 'Unhandled' (in my case and in the examples), so I would prefer to rely on current default mechanism: missing value - missing variable (since there is no statement saying it couldn't be empty)

 - As for error check the users should rely on `statusCode`

docs/modules/plugins/pages/plugin-aws-lambda.adoc
ikalinin1: So, maybe AWS Lambda then ?

valfirst: for S3 we use just S3, also there is no any Lambda which is not AWS

uarlouski (approved):

docs/modules/plugins/pages/plugin-aws-lambda.adoc
valfirst: BTW, the same for: Kafka vs Apache Kafka

ikalinin1: https://www.lambdatest.com/

PR #1191, opened 2020-12-02
[plugin-mobile-app] Fix possible unsupported operation exception on immutable list

uarlouski:


vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/TouchSteps.java
```diff
@@ -128,7 +129,7 @@ public void swipeToElement(SwipeDirection direction, Locator locator, Duration d
     {
         locator.getSearchParameters().setWaitForElement(false);
 
-        List<WebElement> elements = searchActions.findElements(locator);
+        List<WebElement> elements = new ArrayList<>(searchActions.findElements(locator));
```
valfirst, line 132: which part of search actions produces immutable list?

uarlouski: https://github.com/vividus-framework/vividus/blob/master/vividus-extension-selenium/src/main/java/org/vividus/ui/action/search/AbstractElementAction.java there are two places with List.of()

valfirst (approved):

PR #1205, opened 2020-12-07
[xray-exporter] Add missing configuration

uarlouski:


valfirst (commented):
what about adding integrations tests? (like we have for vividus `BeanFactory` and etc.)

valfirst (approved):

ikalinin1 (approved):

PR #1243, opened 2020-12-16
[plugin-saucelabs] Reflect actual result of tests run

TatianaTochko:
SauceLabs job reflects actual result of tests run

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/model/RunningStory.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/model/RunningStory.java
@@ -53,6 +53,11 @@ public class RunningStory
         return failed;
     }
 
+    public void setFailed(boolean failed)
+    {
+        this.failed = failed;
+    }
+
     public boolean isAllowed()
     {
         return allowed;
--- a/vividus-plugin-saucelabs/src/main/java/org/vividus/selenium/sauce/SauceLabsSteps.java
+++ b/vividus-plugin-saucelabs/src/main/java/org/vividus/selenium/sauce/SauceLabsSteps.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


vividus-plugin-saucelabs/src/main/java/org/vividus/selenium/sauce/SauceLabsCapabilitiesConfigurer.java
```diff
@@ -54,6 +61,21 @@ protected SauceConnectOptions createOptions()
         return sauceConnectOptions;
     }
 
+    @AfterScenario
+    public void updateStatusAfterScenario() throws WebDriverException
```
valfirst, line 65: this method should be placed in a separate class

vividus-plugin-saucelabs/src/main/java/org/vividus/selenium/sauce/SauceLabsCapabilitiesConfigurer.java
```diff
@@ -54,6 +61,21 @@ protected SauceConnectOptions createOptions()
...
+    {
+        if (webDriverProvider.isWebDriverInitialized())
```
valfirst, line 67: what if there were several SL sessions within one scenario?

valfirst (changes requested):

vividus-bdd-engine/src/main/java/org/vividus/bdd/model/RunningStory.java
```diff
@@ -53,6 +53,11 @@ public boolean isFailed()
         return failed;
     }
 
+    public void setFailed(boolean failed)
```
valfirst, line 56: we shouldn't add setters just for unit tests, `fail ` method can be used instead

vividus-plugin-saucelabs/src/main/java/org/vividus/selenium/sauce/SauceLabsSteps.java
```diff
@@ -0,0 +1,50 @@
...
+    private IBddRunContext bddRunContext;
+
+    public SauceLabsSteps(IWebDriverProvider webDriverProvider, IBddRunContext bddRunContext)
+    {
+        this.webDriverProvider = webDriverProvider;
+        this.bddRunContext = bddRunContext;
+    }
+
+    @AfterStory
+    public void updateStatusAfterStory() throws WebDriverException
```
valfirst, line 37: it will update the result only if single session is used within whole story, but it's possible to use sessions per scenario, even it's possible to have several sessions within one scenario

valfirst (changes requested):

valfirst:
Superseded by #1329

PR #1263, opened 2020-12-21
[extension-selenium] Do not publish quit event if wd is not initialized

uarlouski:


vividus-extension-selenium/src/main/java/org/vividus/selenium/WebDriverProvider.java
```diff
@@ -84,10 +84,6 @@ public void end()
                 reset(sessionId);
             }
         }
-        else
-        {
-            reset(null);
-        }
     }
 
     private void reset(String sessionId)
```
valfirst, line 89: I think it should be inlined now

valfirst (changes requested):

valfirst (approved):

ikalinin1 (approved):

PR #1268, opened 2020-12-23
[vividus] Add generic while-like step with iteration delays

### LINKED ISSUE #1235: DynamoDB plugin updating
I would like to enhance existing DynamoDB plugin`s functionality and make possibility to set "wait" time while recording appears or update some of its values in case it is already set "by default". ->"Query/GetItem" operation 

Examples:

wait -> add possibility to make "wait period" while file appears (it might be 10-30 seconds or 6-7 minutes for waiting). as example - refresh every 5 second while expected file will appeared
add possibility for waiting when some of attributes of our recording will be changed.
example -> I found recording by "fileName" part (...201207T142421885.xml) and it has "sts"="Pending"
I want to check that "sts" (status) will be changed and appears "sts"="Processed"

DynamoDB plugin => https://github.com/vividus-framework/vividus/issues/1175

valfirst:
Closes #1235

### DIFF
--- a/vividus-tests/src/main/resources/story/integration/ExecutableSteps.story
+++ b/vividus-tests/src/main/resources/story/integration/ExecutableSteps.story
@@ -80,3 +80,12 @@ When I execute steps at most 5 times while variable `var` is < `10`:
 |step                                                                          |
 |When I initialize the scenario variable `var` with value `#{eval(${var} + 1)}`|
 Then `${var}` is = `10`
+
+Scenario: Verify step: When I execute steps with delay `$delay` at most $max times while variable `$variableName` is $comparisonRule `$expectedValue`:$stepsToExeute
+Meta:
+    @requirementId 1235
+When I initialize the scenario variable `var` with value `5`
+When I execute steps with delay `PT0.5S` at most 2 times while variable `var` is < `7`:
+|step                                                                          |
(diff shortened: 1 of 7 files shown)


vividus-tests/src/main/resources/story/integration/ExecutableSteps.story
```diff
@@ -80,3 +80,10 @@ When I execute steps at most 5 times while variable `var` is < `10`:
 |step                                                                          |
 |When I initialize the scenario variable `var` with value `#{eval(${var} + 1)}`|
 Then `${var}` is = `10`
+
+Scenario: Verify step: When I execute steps at most $max times while variable `$variableName` is $comparisonRule `$expectedValue`:$stepsToExeute
```
ikalinin1, line 84: Scenario title differs from the step under test

ikalinin1 (approved):

vividus-tests/src/main/resources/story/system/AWS - DynamoDB.story
```diff
@@ -79,3 +79,18 @@ Then JSON element from `${song}` by JSON path `$` is equal to `
         "SongTitle": "Real Sugar"
     }
 ]` ignoring array order
+
+Scenario: Wait for the data
```
uarlouski, line 83: is ticker id missing?

uarlouski (approved):

PR #1284, opened 2020-12-29
[vividus] Add multisuite support

dKatechev:
- configuration.suites - new property used to set one or more suites separated by comma
- configuration.suite - property is deprecated

### DIFF
--- a/docs/modules/ROOT/pages/tests-configuration.adoc
+++ b/docs/modules/ROOT/pages/tests-configuration.adoc
@@ -2,17 +2,19 @@
 
 == `configuration.properties`
 
-The Vividus configuration includes the following parts: profiles, environments and suite.
+The Vividus configuration includes the following parts: profiles, environments and suites.
 
-=== Suite
+=== Suites
 
-The suite is defined by the property:
+The property `configuration.suites` defines the suites set. It is a comma separated set of suite file addresses:
 
-```properties
-configuration.suite
-```
-
+.configuration.properties
+----
+configuration.suites=testSuite,anotherDir/anotherTestSuite
+----
 The Vividus suite consists of the unlimited number of batches. In its turn the batch consists of any number of stories represented by the files with `*.story` extension. The batches are run sequentially, while the stories in the batch are run in parallel.
+
+The suites are loaded one by one starting from the last one in the sequence, in other words, the values of the properties defined in the first suite will take precedence over the values of the same properties defined in all subsequent suites.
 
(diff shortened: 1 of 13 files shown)


valfirst:
Also please update the regular docs: https://github.com/vividus-framework/vividus/blob/master/docs/modules/ROOT/pages/tests-configuration.adoc (this page is published as https://docs.vividus.dev/vividus/latest/tests-configuration.html)

See https://github.com/vividus-framework/docs.vividus.dev for guidelines on how to generate docs locally

docs/modules/ROOT/pages/tests-configuration.adoc
```diff
@@ -2,18 +2,21 @@
...
+The property `configuration.suites` defines the suites set. It is a comma separated set of suite file addresses:
 
+.configuration.properties
+----
+configuration.suites=testSuite,anotherDir/anotherTestSuite
+----
 The Vividus suite consists of the unlimited number of batches. In its turn the batch consists of any number of stories represented by the files with `*.story` extension. The batches are run sequentially, while the stories in the batch are run in parallel.
 
+Each property in suite files should be unique among suite files provided if you suppose to use them in suites set.
+Otherwise, the behaviour can be unpredictable.
```
valfirst, line 18: I propose to replace this statement with the following one:
```
The suites are loaded one by one starting from the last one in the sequence, in other words, the values of the properties defined in the first suite will take precedence over the values of the same properties defined in all subsequent suites.
```

vividus-tests/src/main/resources/properties/suite/additional/suite.properties
```diff
@@ -0,0 +1,2 @@
+bdd.story-loader.batch-1.resource-location=story/integration/additional
+bdd.story-loader.batch-1.resource-include-patterns=*.story
```
valfirst, line 2:
```properties
bdd.story-loader.batch-4.resource-location=story/integration/additional
bdd.story-loader.batch-4.resource-include-patterns=*.story
```
otherwise the same properties from `integration` suite will override the properties from `additional` suite

valfirst (changes requested):

docs/modules/ROOT/pages/tests-configuration.adoc
dKatechev: Done.

vividus-tests/src/main/resources/properties/suite/additional/suite.properties
dKatechev: Done.

valfirst (approved):

uarlouski (approved):

PR #1324, opened 2021-01-11
Fix docs

ikalinin1:


docs/modules/plugins/pages/plugin-browser-stack.adoc
```diff
@@ -45,7 +45,7 @@ TIP: Links to BrowserStack test runs can be found in Allure report
 
 == Properties
 
-* See xref:tests-development:tests-configuration.adoc#_selenium_grid[Selenium Grid] properties to configure generic options.
+* See xref:ROOT:tests-configuration.adoc#_selenium_grid[Selenium Grid] properties to configure generic options.
```
valfirst, line 48: I think we should refactor the location of Selenium Grid configuration, it's not a common one, it's applicable only for `vividus-plugin-web-app`, `vividus-plugin-mobile-app` and test clouds plugins

valfirst (changes requested):

docs/modules/plugins/pages/plugin-browser-stack.adoc
ikalinin1: Isn't it out of the scope of the pull request? :)

valfirst (approved):

PR #1374, opened 2021-01-21
[bdd-engine] Improve readability of log message about variable saving

valfirst:


vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java
```diff
@@ -50,7 +50,11 @@ public void putVariable(Set<VariableScope> variableScopes, String variableKey, O
     @Override
     public void putVariable(VariableScope variableScope, String variableKey, Object variableValue)
     {
-        LOGGER.info("Saving a value '{}' into the '{}' variable '{}'", variableValue, variableScope, variableKey);
+        LOGGER.atInfo()
+                .addArgument(variableValue)
+                .addArgument(() -> variableScope.name().toLowerCase().replace('_', ' '))
```
ikalinin1, line 55: `variableScope.name().toLowerCase().replace('_', ' ')` why not do it in the enum ?

ikalinin1 (approved):

vividus-bdd-engine/src/main/java/org/vividus/bdd/context/BddVariableContext.java
valfirst: already considered but:
-  override `toString` - may have unknown impact
- new method e.g. `toReadableString` -  this code snippet is used once and I don't see any other places where it could be useful

valfirst: but anyway you gave me an idea

ikalinin1 (approved):

uarlouski (approved):

PR #1386, opened 2021-01-28
[xray-exporter] Take into account scenario statuses in after and before scenario steps

uarlouski:


vividus-to-xray-exporter/src/main/java/org/vividus/xray/factory/TestExecutionFactory.java
```diff
@@ -83,9 +82,11 @@ private static TestExecutionItem createTestInfo(Entry<String, Scenario> scenario
         return test;
     }
 
-    private static TestExecutionItemStatus calculateStatus(List<Step> steps)
+    private static TestExecutionItemStatus calculateStatus(AbstractStepsContainer stepsContainer)
     {
-        return steps.stream().allMatch(step -> "successful".equals(step.getOutcome())) ? TestExecutionItemStatus.PASS
-                : TestExecutionItemStatus.FAIL;
+        return Stream.of(stepsContainer.getSteps(), stepsContainer.getAfterScenarioSteps())
```
valfirst, line 87: why is `BeforeScenarioSteps` not taken into account?

valfirst (changes requested):

vividus-bdd-engine/src/main/java/org/vividus/bdd/model/jbehave/Scenario.java
```diff
@@ -64,9 +58,9 @@ public void setExamples(Examples examples)
 
     public List<Step> collectSteps()
     {
-        if (steps != null)
+        if (getSteps() != null)
         {
-            return steps;
+            return getSteps();
         }
         return Optional.ofNullable(examples)
```
valfirst, line 65: also I believe you should somehow consider before/after scenario steps of Examples

uarlouski: this method is only needed to get steps that go into Xray scenario

uarlouski: can rename it to collectScenarioSteps

valfirst (approved):

PR #1397, opened 2021-02-01
[vividus] Expand Lists with single element at comparison against table

### LINKED ISSUE #1388: Step "Then `$variable` is equal to table ignoring extra columns:$table" fails with an error
**Describe the bug**
Steps:
When I capture HTTP GET or POST request with URL pattern `${event-url-pattern}<event>.*` and save URL query to scenario variable `request-params`
Then `${request-params}` is equal to table ignoring extra columns:
{headerSeparator=!,valueSeparator=!}
!c22    !c52    !
!Download > Other > SZ Caregiver Doctor Discussion Guide !Body !

an error:
issue

Worked fine on VIVIDUS 0.2.12 and earlier.

**To Reproduce**
run steps:
Given I am on a page with the URL '${authority}<relative_url>'
When I click on element located `By.cssSelector(<click-locator>)`
When I wait until HTTP GET or POST request with URL pattern `${event-url-pattern}<event>.*` exists in proxy log
When I capture HTTP GET or POST request with URL pattern `${event-url-pattern}<event>.*` and save URL query to scenario variable `request-params`
Then `${request-params}` is equal to table ignoring extra columns:
{headerSeparator=!,valueSeparator=!}
{headerSeparator=!,valueSeparator=!}
!c22    !c52    !
!Download > Other > SZ Caregiver Doctor Discussion Guide !Body !

**Expected behavior**
No error in the last step

valfirst:
Closes #1388

### DIFF
--- a/vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
+++ b/vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
@@ -16,6 +16,7 @@
 
 package org.vividus.bdd.steps;
 
+import static org.apache.commons.lang3.Validate.isTrue;
 import static org.hamcrest.text.MatchesPattern.matchesPattern;
 
 import java.io.IOException;
@@ -23,6 +24,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.regex.Pattern;
 
@@ -30,6 +32,7 @@ import org.jbehave.core.annotations.Given;
 import org.jbehave.core.annotations.Then;
 import org.jbehave.core.annotations.When;
 import org.jbehave.core.model.ExamplesTable;
+import org.jbehave.core.steps.Parameters;
 import org.vividus.bdd.context.IBddVariableContext;
 import org.vividus.bdd.util.EnumUtils;
 import org.vividus.bdd.util.MapUtils;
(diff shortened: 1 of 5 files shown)


vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
```diff
@@ -177,18 +180,41 @@ public boolean compareVariables(Object variable1, ComparisonRule condition, Obje
...
+    @SuppressWarnings("unchecked")
     public void tablesAreEqualIgnoringExtraColumns(Object variable, ExamplesTable table)
     {
-        if (!(variable instanceof Map))
-        {
-            throw new IllegalArgumentException("'variable' should be instance of map");
-        }
-        Map<String, String> expectedMap = MapUtils.convertSingleRowExamplesTableToMap(table);
-        List<EntryComparisonResult> results = ComparisonUtils.checkMapContainsSubMap((Map<?, ?>) variable, expectedMap);
+        isTrue(variable instanceof Map, "'variable' should be instance of map");
```
uarlouski, line 186: out of scope but maybe The 'variable' ...

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
```diff
@@ -177,18 +180,41 @@ public boolean compareVariables(Object variable1, ComparisonRule condition, Obje
...
+        Map<Object, Object> actualMap = expandListValuesWithSingleElement((Map<Object, Object>) variable);
+
+        List<Parameters> rows = table.getRowsAsParameters(true);
+        isTrue(rows.size() == 1, "ExamplesTable should contain single row with values");
```
uarlouski, line 190: Why not just table? This sounds like a end-user term whereas ExamplesTable is a class name/developer term

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
valfirst: you can consult experts 😄  but AFAIK `'variable'` is a name of the input parameter, so it doesn't require articles

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
```diff
@@ -177,18 +180,41 @@ public boolean compareVariables(Object variable1, ComparisonRule condition, Obje
...
+
+        Map<String, String> expectedMap =  rows.get(0).values();
+        List<EntryComparisonResult> results = ComparisonUtils.checkMapContainsSubMap(actualMap, expectedMap);
         publishMapComparisonResults(List.of(results));
         softAssert.assertTrue(TABLES_ARE_EQUAL, results.stream().allMatch(EntryComparisonResult::isPassed));
     }
 
+    private Map<Object, Object> expandListValuesWithSingleElement(Map<Object, Object> map)
```
uarlouski, line 198: if we have 10 entries in the map where first 9 entries have a value List with size 1, but the last one is different, we will eventually return the map passed as param and put operation on expandedMap will be redundant, so does it make sense to check first that the input map has list with size 1? (for example via allMatch)

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
valfirst: ExamplesTable is a part of JBehave terminology, tcase/whitespaces are discussable, but there should be clear ExamplesTable reference

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
valfirst: in this way it'll be required to traverse the map twice and execute similar logic, if you'd like, you can create a benchmark

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
uarlouski: End user can guess that we use JBehave only by plugin installed in eclipse, should not we advice end users somewhere in the docs to get acquainted with JBehave by giving some links?

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
uarlouski: even if we imagine we have a dataset where the double for is slower it's a bit weird to populate map that is potentially will never be used

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
valfirst: we refer ExamplesTable in many variations in docs, so if you'd like to review the usage of this term you can take care

as for current state of the docs: it's a reference book, it's not an ultimate guide

vividus/src/main/java/org/vividus/bdd/steps/BddVariableSteps.java
valfirst: nothing weird: we're trying to construct a new object, if it doesn't match the rule, we just throw it away
for simple objects: it could be easy to perform the full check and then construct the object, for complex objects: sometimes the check and the iteration could be of higher cost, than creation of object itself, so some flexibility is required

uarlouski (approved):

ikalinin1 (approved):

PR #1526, opened 2021-03-29
[plugin-mobile-app] Add step to press sequence of keys

uarlouski:


vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
```diff
@@ -76,21 +80,68 @@ public void uploadFileToDevice(String filePath)
...
+     * Example:
+     * <br>
+     * <code>
+     * When I press $key key
+     * </code>
      *
-     * @param key key to press
+     * @param key the key to press
      */
     @When("I press $key key")
```
valfirst, line 92: should it be deprecated?

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
```diff
@@ -76,21 +80,68 @@ public void uploadFileToDevice(String filePath)
...
+     * <br>
+     * |Home|
+     * </code>
+     *
+     * @param keys the keys to press
+     */
+    @When("I press keys:$keys")
+    public void pressKeys(ExamplesTable keys)
+    {
+        pressKeys(MapUtils.convertExamplesTableToMap(keys).get("key"));
```
valfirst, line 120: does it make sense to introduce new method in `ExamplesTable`: `getColumn(String key)` ?

valfirst (changes requested):

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: I would prefer to use this step if I need to type only one key, it's 1 vs 3 lines

valfirst: so what about moving it to the composite step in this case?

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
```diff
@@ -76,21 +80,68 @@ public void uploadFileToDevice(String filePath)
...
+                {
+                    AndroidKey androidKey = EnumUtils.getEnumIgnoreCase(AndroidKey.class, key);
+                    Validate.isTrue(androidKey != null, "Unsupported Android key: %s", key);
+                    return androidKey;
+                })
+                .map(KeyEvent::new)
+                .collect(Collectors.collectingAndThen(Collectors.toList(), keyEvents ->
+                {
+                    PressesKey pressesKey = webDriverProvider.getUnwrapped(PressesKey.class);
+                    keyEvents.forEach(pressesKey::pressKey);
```
valfirst, line 142: it could be a part of the wrapping stream

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: End users are expected to use this step extensively, absence of auto-complete or quick docs will hurt their feelings

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: it needs to eliminate multiple invocations of webDriverProvider.getUnwrapped(PressesKey.class)

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
valfirst: not the reason to do not convert it to composite

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
valfirst: just unwrap it once before stream start

uarlouski: what if one of the keys is invalid? then the wrapping becomes pointless

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: why is the end-user unfriendliness not the reason?

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: Added https://github.com/jbehave/jbehave-core/commit/390e2a2bd3abcbe21dc082bc1128af3711b17fd2

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
valfirst: because following this reason all composite steps must be moved to Java

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
valfirst: tricky question: what is better to create a list or to cast an object to another type

vividus-plugin-mobile-app/src/main/java/org/vividus/bdd/mobileapp/steps/DeviceSteps.java
uarlouski: so we don't care about user experience and it's ok for us that the step becomes hardly usable due to the absence of autocomplete and docs

ikalinin1: From my point of view if the step is not deprecated moving into composites will decrease usability, because of absent autocomplete and docs.

I was thinking that composites are for backward compatibility and deprecation, not for "overloading" the steps

ikalinin1 (approved):

valfirst (approved):

PR #1532, opened 2021-04-01
[extension-selenium] Enable detecting of orientation only for mobiles

uarlouski:


ikalinin1 (approved):

vividus-extension-selenium/src/main/java/org/vividus/selenium/manager/GenericWebDriverManager.java
```diff
@@ -262,7 +262,7 @@ protected static Capabilities getCapabilities(WebDriver webDriver)
     @Override
     public boolean isOrientation(ScreenOrientation orientation)
     {
-        if (!isMobile())
+        if (!isMobile() || isTvOS(getCapabilities()))
```
valfirst, line 265: just
```java
if (isMobile())
{
    ScreenOrientation screenOrientation = ...
    ...
}
```

valfirst (changes requested):

vividus-extension-selenium/src/main/java/org/vividus/selenium/manager/GenericWebDriverManager.java
uarlouski: why is it better?

valfirst (approved):

Issue #1460, opened 2021-02-26
Add Azure blob storage v2 support


alex-aladin-a:
Features needed:
1) connect to Azure blob storage
2) list files based on provided initial path
	- with recursive = 'Yes' option to list files in subfolders
	- with recursive = 'No' option to list files in initial folder only
3) filtering feature for listing files:
	- based on file name (+ partial match)
	- based on file name mask
	- based on file extension
	- based on file creation date (<, <=, =, >, => some specific date passed as a parameter)
	- based on file modification date 
	- based on file size
	- based on any combination of above items
4) open file and read the content
5) download a file/folder
6) upload a file/folder

valfirst:
@alex-aladin-a please checkout a new plugin released in [0.3.4](https://github.com/vividus-framework/vividus/releases/tag/release-0.3.4): https://docs.vividus.dev/vividus/0.3.4/plugins/plugin-azure-storage-account.html. It covers many of the requested operations, but not all of them. Feel free to reopen this ticket in case if the required functionality is missed.

PR #1533, opened 2021-04-01
[plugin-azure-storage-account] Introduce plugin for Azure Storage Account

ikalinin1:
Relates #1460


docs/modules/plugins/pages/plugin-azure-storage-blob.adoc
```diff
@@ -0,0 +1,112 @@
+= Azure Storage Blob Plugin
+
+The plugin provides functionality to interact with https://azure.microsoft.com/en-us/services/storage/blobs/[Azure Blob Storage]
+
+== Installation
+
+.build.gradle
+[source,gradle,subs="attributes+"]
+----
+implementation(group: 'org.vividus', name: 'vividus-plugin-azure-storage-blob', version: '{current-version}')
```
valfirst, line 10: I think we should stick to the naming of Azure services:
`vividus-plugin-azure-storage-account`

Blob storage is a part of storage account functionality

docs/modules/plugins/pages/plugin-azure-storage-blob.adoc
```diff
@@ -0,0 +1,112 @@
...
+----
+azure.storage.blob.account.{account-name}.{property-name}=property value
+----
+
+Where:
+
+. `account-name` - The name of the storage account
+. `property-name` - The name of the account property. One of: 
+.. `name` - The name of the storage account
+.. `sas-token` - The https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview[SAS token]
```
valfirst, line 29: Why not to use Azure Active Directory as for other services? SAS token creation could be challenging due to permission restrictions

docs/modules/plugins/pages/plugin-azure-storage-blob.adoc
```diff
@@ -0,0 +1,112 @@
...
+.. `container` - Comma-separated list of containers
```
valfirst, line 30: different storage accounts may have containers with same names

valfirst (changes requested):

uarlouski (approved):

PR #1567, opened 2021-04-16
[CI] Debug snapshots publishing

valfirst:


valfirst:
The issues is resolved, the artifacts are successfully published to Artifactory

PR #1628, opened 2021-05-13
Fix typo

ikalinin1:


docs/modules/plugins/pages/plugin-web-app.adoc
```diff
@@ -244,6 +244,46 @@ When I open URL `https://delayed.vividus.dev` in new window
...
+.Scroll login to a bottom
+[source,gherkin]
+----
+When I change context to element located `id(login)`
+When I scroll context to BOTTOM edge
+----
+
+=== Scroll element into view
+
+Scrolls an element into the view using https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView[JavaScript]
```
uarlouski, line 273: "JavaScript" doesn't give much context without navigating to the referred page, I mean even a man who knows js will have to navigate instead of just getting this info from link text

ikalinin1: Any ideas on how to improve?

valfirst (approved):

docs/modules/plugins/pages/plugin-web-app.adoc
valfirst: I would remove `using https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView[JavaScript]`, because these are details of the implementation

uarlouski (approved):

valfirst (approved):

PR #1633, opened 2021-05-14
[CI] Speed-up workflow execution

ikalinin1:


.github/workflows/gradle.yml
```diff
@@ -190,7 +190,7 @@ jobs:
         fi
 
     - name: SauceLabs - Desktop - Web - System tests
-      if: matrix.platform == 'ubuntu-latest'
+      if: matrix.platform == 'windows-latest'
```
valfirst, line 193: we can't change it in this way: if tests are failed, the build shouldn't be published, but if we move it to `windows`, than builds will be published even in case of failures on `windows` platform

PR #1648, opened 2021-05-18
[plugin-azure-storage-account] Enable usage of blob service SAS URLs

valfirst:


vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -59,7 +59,7 @@
                 @Override
                 public BlobServiceClient load(String endpoint)
                 {
-                    return new BlobServiceClientBuilder().endpoint(endpoint).credential(credential).buildClient();
+                    return new BlobServiceClientBuilder().credential(credential).endpoint(endpoint).buildClient();
```
ikalinin1, line 62: What is the reason for re-ordering here?

ikalinin1 (approved):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
valfirst: 1.
Screen Shot 2021-05-19 at 9 58 40 AM
2.
Screen Shot 2021-05-19 at 9 59 16 AM
3.
Screen Shot 2021-05-19 at 9 59 34 AM

uarlouski (approved):

PR #1651, opened 2021-05-19
[bdd-engine] Fix invalid exceptions thrown from sub steps

### LINKED ISSUE #1564: Error message in Allure report about step deletion is not obvious
Step When I switch to a frame by the xpath '$xpath' was depricated and removed starting from Vividus version 0.3.0 and replaced with When I switch to frame located `$locator`.

Using the old step on version 0.3.0 brakes scenarios, but the error message about the failure root cause is not obvious.

Message update is needed.
[logs.txt](https://github.com/vividus-framework/vividus/files/6310392/logs.txt)

uarlouski:
Closes #1564

### DIFF
--- a/vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/SubSteps.java
+++ b/vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/SubSteps.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2020 the original author or authors.
+ * Copyright 2019-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/SubSteps.java
```diff
@@ -88,6 +88,10 @@ else if (rootCause instanceof Error)
         {
             throw (Error) rootCause;
         }
+        else if (rootCause instanceof UUIDExceptionWrapper)
+        {
+            throw new UUIDExceptionWrapper(rootCause);
```
valfirst, line 93: double-wrapping?

uarlouski: - https://github.com/jbehave/jbehave-core/blob/master/jbehave-core/src/main/java/org/jbehave/core/steps/StepCreator.java#L889
- and as per unit tests
- technically this is a single wrapping
Screenshot 2021-05-19 at 11 59 14

valfirst (approved):

vividus-bdd-engine/src/main/java/org/vividus/bdd/steps/SubSteps.java
valfirst: `throw stepResult.getFailure();` ?

valfirst (changes requested):

valfirst (approved):

ikalinin1 (approved):

PR #1763, opened 2021-07-01
Add ability to compare for `null` values in ExamplesTable

### LINKED ISSUE #1583: Ability to assert null values from Postgresql DB
If null values appear in a select from a postgresql it cannot be compared correctly to an Examples table from test scenario. An assertion fails because "null" has null data type in the select and text data type in the test scenario.
image

It would be great to have a possibility to assert "null" values from DB to assert full select without  any exclusions.

Please note that a types transformation to text during select is not working for null values. Assertion fails anyway.

valfirst:
Closes #1583

### DIFF
--- a/vividus-plugin-db/src/main/java/org/vividus/bdd/steps/db/DatabaseSteps.java
+++ b/vividus-plugin-db/src/main/java/org/vividus/bdd/steps/db/DatabaseSteps.java
@@ -21,6 +21,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


uarlouski (approved):

vividus-plugin-db/src/main/java/org/vividus/bdd/steps/db/DatabaseSteps.java
```diff
@@ -409,7 +410,9 @@ public void compareData(List<Map<String, Object>> leftData, Set<String> keys, St
 
     private Map<String, Object> convertValuesToString(Map<String, Object> map)
     {
-        return map.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, e -> String.valueOf(e.getValue())));
+        Map<String, Object> result = new HashMap<>();
+        map.forEach((key, value) -> result.put(key, value == null ? null : String.valueOf(value)));
```
ikalinin1, line 414: since we sure it is not null `value.toString()` to avoid double checks?

ikalinin1 (approved):

vividus-plugin-db/src/main/java/org/vividus/bdd/steps/db/DatabaseSteps.java
ikalinin1: To avoid double checks

PR #1771, opened 2021-07-06
Update SECURE_KEY_PATTERN for Logger to hide <private-key> properties

tshkursk:


vividus/src/main/java/org/vividus/report/MetadataLogger.java
```diff
@@ -54,7 +54,7 @@
     private static final String NEW_LINE = "%n";
     private static final Logger LOGGER = LoggerFactory.getLogger(MetadataLogger.class);
     private static final Pattern SECURE_KEY_PATTERN = Pattern
-            .compile(".*(password|access-?key|api-?key|secret|token).*", Pattern.CASE_INSENSITIVE);
+            .compile(".*(password|access-?key|api-?key|secret|token|private-?key).*", Pattern.CASE_INSENSITIVE);
```
ikalinin1, line 57: what about `key` ?

tshkursk: I can add if really needed. Please, provide usage examples?

ikalinin1: The idea is to have a generic matcher for anything with the `key` part

tshkursk: The question is do you really want to hide everithing with `key` part?
Looks like lot of confings could be affected, e.g.: https://docs.vividus.dev/vividus/latest/plugins/plugin-azure-cosmos-db.html#_configuration

valfirst, line 57:
```suggestion
            .compile(".*(password|((access|api|private)-)?(key|secret|token)).*" Pattern.CASE_INSENSITIVE);
```

valfirst (changes requested):

ikalinin1 (approved):

valfirst (approved):

PR #1850, opened 2021-08-04
[tests] Switch system saucelabs Android tests to Appium 1.20.2

ikalinin1:


abudevich (approved):

ikalinin1:
> I think it would be 0.4.0 only

But its tests update only

uarlouski (approved):

vividus-tests/src/main/resources/properties/environment/system/saucelabs/mobile_app/ios/environment.properties
```diff
@@ -1,6 +1,6 @@
-selenium.grid.appium-version=1.21.0
+selenium.grid.appium-version=1.22.0
```
valfirst, line 1: I would move this update to a separate commit and upgraded the version for Android system tests as well

valfirst (approved):

uarlouski (approved):

PR #1864, opened 2021-08-12
[plugin-azure-storage-account] Add step to collect storage accounts info

valfirst:


vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/StorageAccountManagementSteps.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+        };
+        this.jacksonAdapter = new JacksonAdapter()
+        {
+            private ObjectMapper objectMapper;
+
+            @Override
+            protected ObjectMapper simpleMapper()
+            {
+                if (this.objectMapper == null)
```
uarlouski, line 68: in which case this.objectMapper != null here?

valfirst: azure-jackson

uarlouski (approved):

PR #1877, opened 2021-08-16
Add step to collect system topics info

valfirst:


vividus-extension-azure/src/main/java/org/vividus/azure/util/InnersJacksonAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.introspect.Annotated;
+import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
+
+public class InnersJacksonAdapter extends JacksonAdapter
+{
+    private ObjectMapper objectMapper;
+
+    @Override
+    protected ObjectMapper simpleMapper()
```
uarlouski, line 30: is it ok that OM initialization is not thread-safe?

uarlouski (approved):

ikalinin1 (approved):

PR #1885, opened 2021-08-18
[util] Update to use non-deprecated Apache Commons Pool API

valfirst:


vividus-util/src/main/java/org/vividus/util/pool/UnsafeGenericObjectPool.java
```diff
@@ -25,12 +26,12 @@
 
 public class UnsafeGenericObjectPool<T> extends GenericObjectPool<T>
```
ikalinin1, line 27: Can we just drop this logic instead? :)

ikalinin1 (approved):

uarlouski (approved):

PR #1948, opened 2021-09-08
[bdd-engine] Add ability to load ExamplesTable bodies from variable

valfirst:


vividus-tests/src/main/resources/properties/suite/integration/suite.properties
```diff
@@ -12,6 +12,8 @@ bdd.batch-1.meta-filters=groovy: !skip && epic != 'vividus-plugin-ssh' && !local
 bdd.batch-1.threads=5
 bdd.batch-1.story-execution-timeout=PT10M
 bdd.variables.batch-1.locale=ca
+bdd.variables.batch-1.filtered-examples-table={transformer=FILTERING, byColumnNames=column3}\
```
ikalinin1, line 15: does it make sense to put some variables into this table definition? global/scoped?

vividus-bdd-engine/src/main/java/org/jbehave/core/model/VariableResolvingTableParsers.java
```diff
@@ -0,0 +1,42 @@
...
+package org.jbehave.core.model;
+
+import java.util.Optional;
+
+import org.jbehave.core.configuration.Keywords;
+import org.jbehave.core.model.ExamplesTable.TablePropertiesQueue;
+import org.jbehave.core.steps.ParameterConverters;
+import org.vividus.bdd.steps.VariableResolver;
+
+public class VariableResolvingTableParsers extends TableParsers
```
ikalinin1, line 26: Please add unit tests for the class

PR #1966, opened 2021-09-15
[plugin-mobile-app] Fix staleness exepction during text typing on iOS

ikalinin1:


uarlouski (approved):

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/action/KeyboardActions.java
```diff
@@ -127,4 +128,18 @@ else if (realDevice)
...
+        try
+        {
+            tagName = webElement.getTagName();
+        }
+        catch (StaleElementReferenceException exception)
+        {
+            // Swallow exception quietly
+        }
+        return tagName;
+    }
```
valfirst, line 144:
```suggestion
    private String getTagNameSafely(WebElement webElement)
    {
        try
        {
            return webElement.getTagName();
        }
        catch (StaleElementReferenceException exception)
        {
            // Swallow exception quietly
            return null;
        }
    }
```

valfirst (changes requested):

valfirst (approved):

PR #1972, opened 2021-09-17
Azure storage account steps

tshkursk:


vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                           <li><code>maxNumber</code> Optional. Maximum number of blobs per page.</li>
+     *                          </ul>
+     */
+    @When("I find blobs and save result to $scopes variable `$variableName` with parameters:$searchParameters")
+    public void findBlobsByParameters(Set<VariableScope> scopes, String variableName, ExamplesTable searchParameters)
+    {
+        List<String> blobNames = new ArrayList<>();
+        for (Parameters row : searchParameters.getRowsAsParameters(true))
+        {
+            StringComparisonRule rule = row.valueAs("rule", StringComparisonRule.class);
```
ikalinin1, line 275: Please consider moving these to the POJO

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                         <li><b>STEP</b> - the variable will be available only within the step,
+     *                         <li><b>SCENARIO</b> - the variable will be available only within the scenario,
+     *                         <li><b>STORY</b> - the variable will be available within the whole story,
+     *                         <li><b>NEXT_BATCHES</b> - the variable will be available starting from next batch
+     *                         </ul>
+     * @param variableName     The variable name to store the list of found blob names.
+     * @param searchParameters Parameters to search blob names.
+     *                         <div>Example:</div>
+     *                         <code>
+     *                         <br>When I find blobs and save result to SCENARIO variable `blobs` with parameters::
```
abudevich, line 253: extra `:`

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+            {
+                options.setMaxResultsPerPage(maxNumber);
+            }
+            BlobContainerClient blobContainerClient = createBlobContainerClient(container, storageAccountKey);
+            Matcher<String> nameMatcher = rule.createMatcher(blobNameToMatch);
+            blobNames.addAll(blobContainerClient.listBlobsByHierarchy("/", options, null)
+                    .stream()
+                    .map(BlobItem::getName)
+                    .filter(nameMatcher::matches)
+                    .collect(Collectors.toList()));
```
ikalinin1, line 292: .forEach(blobNames::add) ?

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                         <i>Available scopes:</i>
+     *                         <ul>
+     *                         <li><b>STEP</b> - the variable will be available only within the step,
+     *                         <li><b>SCENARIO</b> - the variable will be available only within the scenario,
+     *                         <li><b>STORY</b> - the variable will be available within the whole story,
+     *                         <li><b>NEXT_BATCHES</b> - the variable will be available starting from next batch
+     *                         </ul>
+     * @param variableName     The variable name to store the list of found blob names.
+     * @param searchParameters Parameters to search blob names.
+     *                         <div>Example:</div>
```
ikalinin1, line 251: Could you please elaborate what is the reason to wrap it into <div></div> ?

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -135,7 +139,7 @@ public void downloadBlobToFile(String blobName, String containerName, String sto
             Set<VariableScope> scopes, String variableName) throws IOException
     {
         String tempFilePath = ResourceUtils.createTempFile(baseFileName).toAbsolutePath().toString();
-        createBlobClient(blobName, containerName, storageAccountKey).downloadToFile(tempFilePath);
+        createBlobClient(blobName, containerName, storageAccountKey).downloadToFile(tempFilePath, true);
```
valfirst, line 142: I would rather avoid creating the file in the previous line

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                         <br>
+     *                         <br>where
+     *                         <ul>
+     *                           <li><code>rule</code> The blob name comparison rule: "is equal to", "contains",
+     *                           "does not contain" or "matches".</li>
+     *                           <li><code>blobNameToMatch</code> The full or partial blob name to be matched.</li>
+     *                           <li><code>container</code> The name of the container to point to.</li>
+     *                           <li><code>storageAccountKey</code> The key to Storage Account endpoint.</li>
+     *                           <li><code>directory</code> The directory to list blobs underneath.</li>
+     *                           <li><code>maxNumber</code> Optional. Maximum number of blobs per page.</li>
```
valfirst, line 266: not clear, there is no pagination in this step

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
valfirst, line 253: I recommend to use a linear syntax (without ExamplesTable)

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                         </code>
+     *                         <br>
+     *                         <br>where
+     *                         <ul>
+     *                           <li><code>rule</code> The blob name comparison rule: "is equal to", "contains",
+     *                           "does not contain" or "matches".</li>
+     *                           <li><code>blobNameToMatch</code> The full or partial blob name to be matched.</li>
+     *                           <li><code>container</code> The name of the container to point to.</li>
+     *                           <li><code>storageAccountKey</code> The key to Storage Account endpoint.</li>
+     *                           <li><code>directory</code> The directory to list blobs underneath.</li>
```
valfirst, line 265: it's not clear how `rule` + `blobNameToMatch` works in conjunction with `directory`

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +235,65 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+    {
+        List<String> blobNames = new ArrayList<>();
+        for (Parameters row : searchParameters.getRowsAsParameters(true))
+        {
+            StringComparisonRule rule = row.valueAs("rule", StringComparisonRule.class);
+            String blobNameToMatch = row.valueAs("blobNameToMatch", String.class);
+            String container = row.valueAs("container", String.class);
+            String storageAccountKey = row.valueAs("storageAccountKey", String.class);
+            String directory = row.valueAs("directory", String.class);
+            ListBlobsOptions options = new ListBlobsOptions().setPrefix(directory);
```
valfirst, line 280: it seems it's not a true `directory`, but just prefix (I assume if there is a virtual directory: `/abc/def`, then it will work for both: `/abc` and `abc/d`)

valfirst (commented):
Please add docs for new step

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
valfirst: please disregard this comment, now Java always creates an empty file (no easy way to generate the path to temp file without its creation)

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
 When I find all blobs with name which contains `json` in container `vividus` of storage account `testaccount` and save result to scenario variable `blobsCollection`
 ----
 
+==== Find blobs by filters
+
+Find blobs with names filtered by specified filters in the container.
```
valfirst, line 174:
```suggestion
Find blobs with names filtered by the specified rules in the container.
```

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
...
+All filters are optional.
+
+[source,gherkin]
+----
+When I filter blobs by:$filters in container `$containerName` of storage account `$storageAccountKey` and save result to $scopes variable `$variableName`
+----
+
+* `$filters` Table with filters to apply on blob names.
```
valfirst, line 182:
```suggestion
* `$filters` - Table with filters to apply on blob names.
```

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
...
+.The supported filters
+[cols="1,1,2", options="header"]
+|===
+|Filter
+|Description
+|Example value
+
+|`blobNamePrefix`
+|The directory(prefix) to list blobs underneath.
+|path/to/data/
```
valfirst, line 193: is root slash required or not? e.g. is `/path/to/data` equal to `path/to/data`?

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
...
++
+.The supported filters
+[cols="1,1,2", options="header"]
+|===
+|Filter
+|Description
+|Example value
+
+|`blobNamePrefix`
+|The directory(prefix) to list blobs underneath.
```
valfirst, line 192:
```suggestion
|The prefix which blob names should start with
```

we can't use `directory` term for any blob storage

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
         bddVariableContext.putVariable(scopes, variableName, blobNames);
     }
 
+    /**
+     * Finds blobs with names filtered by specified filters in the container.
```
valfirst, line 240:
```suggestion
     * Finds blobs with names filtered by the specified rules in the container.
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *
+     * @param filters           Filters to apply on blob names.
```
valfirst, line 242:
```suggestion
     * @param filters           Filters to apply to blob names.
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                          </code>
+     *                          <br>
+     *                          <br>where all filters are optional
+     *                          <ul>
+     *                           <li><code>blobNameFilterRule</code> The blob name comparison rule: "is equal to",
+     *                           "contains", "does not contain" or "matches".
+     *                           Should be specified along with <i>blobNameFilterValue</i>.</li>
+     *                           <li><code>blobNameFilterValue</code> The full or partial blob name to be matched.
+     *                           Should be specified along with <i>blobNameFilterRule</i>.</li>
+     *                           <li><code>blobNamePrefix</code> The directory(prefix) to list blobs underneath.</li>
```
valfirst, line 260:
```suggestion
     *                           <li><code>blobNamePrefix</code> The prefix which blob names should start with.</li>
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     */
+    @When("I filter blobs by:$filters in container `$containerName` of storage account `$storageAccountKey`"
+            + " and save result to $scopes variable `$variableName`")
+    public void findBlobsByFilters(List<BlobFilter> filters, String containerName,
+                                      String storageAccountKey, Set<VariableScope> scopes, String variableName)
+    {
+        List<BlobFilter> checkedFilters = checkFilters(filters);
+        List<String> blobNames = new ArrayList<>();
+        BlobContainerClient blobContainerClient = createBlobContainerClient(containerName, storageAccountKey);
+        for (BlobFilter filter : checkedFilters)
```
valfirst, line 284: tricky case: is it required to support multiple filters?

what if one filter conflicts with another one? or vice versa: they produce the intersecting result sets?

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+        {
+            blobContainerClient
+                    .listBlobsByHierarchy(filter.getBlobNamePrefix().orElse(StringUtils.EMPTY))
```
valfirst, line 287: why is an empty string used as default value? we shouldn't set this filter if no option is specified to avoid relying Azure SDK defaults

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+                    .stream()
+                    .map(BlobItem::getName)
+                    .filter(filterMatched(filter))
+                    .limit(filter.getResultsLimit().orElse(Long.MAX_VALUE))
```
valfirst, line 291: this approach won't work if only limit is specified: we should rely on server-side limit

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+                filter.setBlobNameFilterValue(Optional.empty());
+            }
+            if (filter.getBlobNamePrefix() == null)
+            {
+                filter.setBlobNamePrefix(Optional.empty());
+            }
+            if (filter.getResultsLimit() == null)
+            {
+                filter.setResultsLimit(Optional.empty());
+            }
```
valfirst, line 315: do not change the state of input parameters, it can be done via implementation of addition POJO API for convenience

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +236,100 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+            if (filter.getBlobNameFilterValue().isPresent() && filter.getBlobNameFilterRule().isEmpty()
+                    || filter.getBlobNameFilterValue().isEmpty() && filter.getBlobNameFilterRule().isPresent())
+            {
+                throw new IllegalArgumentException(
+                        "Should be specified together: 'blobNameFilterRle' and 'blobNameFilterValue'");
```
valfirst, line 320: the same as above: move validation to POJO

valfirst (changes requested):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
tshkursk: Server-side limit is `maxResultsPerPage` - but, as discussed with Anton, they don't need limit `per page`, but global one.

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
tshkursk: I guess the profit of step is to get multiple blobs from different directories with one step-call. Each filter is applied on initial container, not on previous filtered set. Filters are independent, they just summarize names in result set.

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
valfirst: because the goal of the limit is to reduce the number of results: if the limit is applied on the client side, the client will download all available results at first, and after then limit will be applied - that's performance bottleneck you are trying to eliminate

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
valfirst: 1) there is high risk to get duplicate results
2) the overall complexity of the step is growing: all rules should be documented and explained to end-users
3) do you have any example when 2 and more filters are needed?

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
tshkursk: I've re-checked on target huge container - no performance problems with current limit.

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jbehave.core.annotations.AsParameters;
+import org.vividus.bdd.steps.StringComparisonRule;
+
+@AsParameters
+public class BlobFilter
+{
+    private String blobNamePrefix;
+    private StringComparisonRule blobNameFilterRule;
+    private String blobNameFilterValue;
+    private Long resultsLimit;
```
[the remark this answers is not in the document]
tshkursk: `Optional` was actual when I have multiple filters in a table. Like this:
```
When I filter blobs by:
| blobNamePrefix  | blobNameFilterRule | blobNameFilterValue |
| data1/                 |                                  |                                    |
| data2/                 | contains                    | file-key.txt                  |
in container `$containerName` of storage account `$storageAccountKey` and save result to $scopes variable `$variableName`
```
 If I use Optional<blobNamePrefix> - this mean I can use step with empty <blobNamePrefix> filter - not clear for what.
```
When I filter blobs by:
| blobNamePrefix  |
|                             |
in container `$containerName` of storage account `$storageAccountKey` and save result to $scopes variable `$variableName`
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +238,112 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+                        .forEach(resultBlobNames::add);
+                }
+        );
+        return resultBlobNames;
+    }
+
+    private static List<String> getLimitPerPage(BlobFilter filter, Long limit, PagedIterable<BlobItem> pages)
+    {
+        List<String> blobNames = new ArrayList<>();
+        for (PagedResponse<BlobItem> page: pages.iterableByPage())
```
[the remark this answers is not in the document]
tshkursk: Didn't find appropriate example how to use it here. Please, explain or show an example?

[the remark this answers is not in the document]
valfirst: you can test it: just make sure that at least the second page is returned as expected

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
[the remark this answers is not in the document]
valfirst: > If I use Optional - this mean I can use step with empty filter - not clear for what.

that's the matter of validation

in general the goal of `Optional` here is to make API clearer, because `Optional` is a clear indication whether field is nullable or not

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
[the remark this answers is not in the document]
tshkursk: https://github.com/vividus-framework/vividus/pull/1972/files#diff-cc3401e9cf8e54d142ec949914319323d2ed8fcb61355b8f1d0db04dac65a34cR257
here's how I covered this

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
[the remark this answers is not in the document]
tshkursk: So I guess we don't need nullable filed here. I just either specify needed column or not.

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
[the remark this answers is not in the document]
valfirst: I meant to check the behavior of the step in the real blob storage.
The stubbing of invocations in unit tests doesn't guarantee Azure API works in the same way

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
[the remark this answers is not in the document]
valfirst: All invocations in the step like `Optional.ofNullable(filter.getBlobNamePrefix())` shows that such fields are nullable, it means that they should be `Optional<..>` in usual case, but I haven't checked here whether conversion from `ExamplesTable` to `Optional<..>` field works properly

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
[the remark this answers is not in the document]
tshkursk: Checked on real storage. Info logging added to view pagination on test run.

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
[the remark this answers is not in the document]
tshkursk: Changed POJO to avoid `Optional.ofNullable`.

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
...
+|The full or partial blob name to be matched.
+Should be specified along with ```blobNameFilterRule```.
+|`.*`
+|===
+
+* `$containerName` - The name of the container to point to.
+* `$storageAccountKey` - The key of storage account from the configuration.
+* `$scopes` - xref:commons:variables.adoc#_scopes[The comma-separated set of the variables scopes].
+* `$variableName` - The variable name to store the list of found blob names.
+
```
ikalinin1, line 214: Please add usage example

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+                        .map(BlobItem::getName)
+                        .filter(filterMatched(filter))
+                        .collect(Collectors.toList())
+                );
+    }
+
+    private static List<String> getLimitPerPage(BlobFilter filter, Long limit, PagedIterable<BlobItem> pages)
+    {
+        List<String> blobNames = new ArrayList<>();
+        for (PagedResponse<BlobItem> page: pages.iterableByPage())
```
ikalinin1, line 314:
```suggestion
        for (PagedResponse<BlobItem> page : pages.iterableByPage())
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public void setResultsLimit(Optional<Long> resultsLimit)
+    {
+        this.resultsLimit = resultsLimit;
+    }
+
+    public void validate()
+    {
+        isTrue(getResultsLimit() != null && getResultsLimit().isPresent()
```
ikalinin1, line 76: Won't setting empty optional by default simplify this logic?

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,100 @@
...
+                        || getBlobNameFilterValue() != null && getBlobNameFilterValue().isPresent()
+                        || getBlobNameFilterRule() != null && getBlobNameFilterRule().isPresent()
+                        || getBlobNamePrefix() != null && getBlobNamePrefix().isPresent(),
+                "At least one filter is present");
+        isTrue(getBlobNameFilterValue() == null && getBlobNameFilterRule() == null
+                || getBlobNameFilterValue() != null && getBlobNameFilterRule() != null
+                        && (getBlobNameFilterValue().isPresent() && getBlobNameFilterRule().isPresent()
+                        || getBlobNameFilterValue().isEmpty() && getBlobNameFilterRule().isEmpty()),
+                "Should be specified together: 'blobNameFilterRle' and 'blobNameFilterValue'");
```
ikalinin1, line 85: https://github.com/vividus-framework/vividus/pull/1972/files#r744469118

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,49 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
 When I find all blobs with name which contains `json` in container `vividus` of storage account `testaccount` and save result to scenario variable `blobsCollection`
 ----
 
+==== Find blobs by filter
+
+Find blobs with names filtered by the specified rules in the container.
+All rules are optional.
```
uarlouski, line 175: Please add that at least one rule is required

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public void validate()
+    {
+        isTrue(getResultsLimit().isPresent() || getBlobNameFilterValue().isPresent()
+                        || getBlobNameFilterRule().isPresent() || getBlobNamePrefix().isPresent(),
+                "At least one filter is present");
+        isTrue(getBlobNameFilterValue().isPresent() && getBlobNameFilterRule().isPresent()
+                        || getBlobNameFilterValue().isEmpty() && getBlobNameFilterRule().isEmpty(),
+                "Should be specified together: 'blobNameFilterRle' and 'blobNameFilterValue'");
```
ikalinin1, line 81:
```suggestion
        isTrue(getResultsLimit().isPresent() || getBlobNameFilterRule().isPresent() || getBlobNamePrefix().isPresent(),
                "At least one filter is present");
        isTrue(getBlobNameFilterRule().isPresent() ? getBlobNameFilterValue().isPresent() : true
                "Should be specified together: 'blobNameFilterRle' and 'blobNameFilterValue'");
```

tshkursk: Will not consider the case when user specify only BlobNameFilterValue (without BlobNameFilterRule). Related unit test:
https://github.com/vividus-framework/vividus/pull/1972/files#diff-cc3401e9cf8e54d142ec949914319323d2ed8fcb61355b8f1d0db04dac65a34cR302

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                          <div>Example:</div>
+     *                          <code>
+     *                          <br>When I filter blobs by:
+     *                          <br>|blobNamePrefix|blobNameFilterRule|blobNameFilterValue|resultsLimit|
+     *                          <br>|data/         |contains          |file-key.txt       |10          |
+     *                          <br>in container `global` of storage account `storage`
+     *                          <br> and save result to story variable `blobs`
+     *                          </code>
+     *                          <br>
+     *                          <br>where all filters are optional
```
valfirst, line 253:
```suggestion
     *                          <br>where all filters are optional, but at least one rule is required.
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,83 @@
...
+    public void setResultsLimit(Optional<Long> resultsLimit)
+    {
+        this.resultsLimit = resultsLimit;
+    }
+
+    public void validate()
+    {
+        isTrue(getResultsLimit().isPresent() || getBlobNameFilterValue().isPresent()
+                        || getBlobNameFilterRule().isPresent() || getBlobNamePrefix().isPresent(),
+                "At least one filter is present");
```
valfirst, line 78:
```suggestion
                "At least one filter must be present");
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
valfirst, line 81:
```suggestion
        isTrue(getBlobNameFilterValue().isPresent() == getBlobNameFilterRule().isPresent(),
                "Should be specified together: 'blobNameFilterRle' and 'blobNameFilterValue'");
```

vividus-plugin-azure-storage-account/src/test/java/org/vividus/azure/storage/blob/BlobStorageStepsTests.java
```diff
@@ -180,13 +194,133 @@ void shouldListFilteredBlobs()
...
+                    },
+                    () ->  {
+                        PagedIterable<BlobItem> iterable = mock(PagedIterable.class);
+                        when(iterable.stream()).thenReturn(Stream.of(first, second, third));
+                        prefix.ifPresentOrElse(
+                                p -> when(client.listBlobsByHierarchy(eq(p))).thenReturn(iterable),
+                                () -> when(client.listBlobs()).thenReturn(iterable)
+                        );
+                    }
+            );
```
valfirst, line 283: unit tests shouldn't contain so many `if`-logic (aka code branches), in the ideal state there should be no branches at all

valfirst (changes requested):

valfirst (approved):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                          <br>When I filter blobs by:
+     *                          <br>|blobNamePrefix|blobNameFilterRule|blobNameFilterValue|resultsLimit|
+     *                          <br>|data/         |contains          |file-key.txt       |10          |
+     *                          <br>in container `global` of storage account `storage`
+     *                          <br> and save result to story variable `blobs`
+     *                          </code>
+     *                          <br>
+     *                          <br>where all filters are optional, but at least one rule is required.
+     *                          <ul>
+     *                           <li><code>blobNameFilterRule</code> The blob name comparison rule: "is equal to",
```
ikalinin1, line 255: one-space indentation looks confusing

ikalinin1 (approved):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                           "contains", "does not contain" or "matches".
+     *                           Should be specified along with <i>blobNameFilterValue</i>.</li>
+     *                           <li><code>blobNameFilterValue</code> The full or partial blob name to be matched.
+     *                           Should be specified along with <i>blobNameFilterRule</i>.</li>
+     *                           <li><code>blobNamePrefix</code> The prefix which blob names should start with.</li>
+     *                           <li><code>resultsLimit</code> Maximum number of blob names to return.</li>
```
valfirst, line 261:
```suggestion
     *                          <li><code>blobNameFilterRule</code> The blob name comparison rule: "is equal to",
     *                           "contains", "does not contain" or "matches".
     *                           Should be specified along with <i>blobNameFilterValue</i>.</li>
     *                          <li><code>blobNameFilterValue</code> The full or partial blob name to be matched.
     *                           Should be specified along with <i>blobNameFilterRule</i>.</li>
     *                          <li><code>blobNamePrefix</code> The prefix which blob names should start with.</li>
     *                          <li><code>resultsLimit</code> Maximum number of blob names to return.</li>
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+                        .stream()
+                        .map(BlobItem::getName)
+                        .filter(filterMatched(filter))
+                        .collect(Collectors.toList())
+                );
+    }
+
+    private static List<String> getLimitPerPage(BlobFilter filter, Long limit, PagedIterable<BlobItem> pages)
+    {
+        List<String> blobNames = new ArrayList<>();
```
ikalinin1, line 313: If you know the limit I think it makes sense to create fixed-size list?

ikalinin1, line 313: Please create fixed-capacity list

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+        for (PagedResponse<BlobItem> page : pages.iterableByPage())
+        {
+            if (blobNames.size() < limit)
+            {
+                LOGGER.info("Blob Items on page: {}", page.getContinuationToken());
+                page.getValue()
+                        .stream()
+                        .map(BlobItem::getName)
+                        .filter(filterMatched(filter))
+                        .forEach(blobNames::add);
```
ikalinin1, line 323: In your realization you will add the items into the list and then remove them, I propose to rewrite logic to not add them when the limit is reached

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+            }
+            else
+            {
+                break;
+            }
+        }
+        return blobNames.stream().limit(limit).collect(Collectors.toList());
```
ikalinin1, line 330: Please use java.util.List.subList(int fromIndex, int toIndex)

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+    }
+
+    private static Predicate<String> filterMatched(BlobFilter filter)
+    {
+        return name -> filter.getBlobNameFilterValue()
+                    .map(value -> filter.getBlobNameFilterRule().get().createMatcher(value).matches(name))
```
ikalinin1, line 336: I propose to avoid matcher creation for-each blobName to match

ikalinin1, line 336:
```suggestion
        Matcher<String> nameFilter = filter.getBlobNameFilterValue()
                    .map(value -> filter.getBlobNameFilterRule().get().createMatcher(value)
        return name -> nameFilter.matches(name));
```

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +237,106 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+        BlobContainerClient blobContainerClient = createBlobContainerClient(containerName, storageAccountKey);
+        bddVariableContext.putVariable(scopes, variableName, getLimitedBlobNames(blobContainerClient, filter));
+    }
+
+    private static List<String> getLimitedBlobNames(BlobContainerClient blobContainerClient, BlobFilter filter)
+    {
+        return filter.getResultsLimit()
+                .map(limit -> {
+                    ListBlobsOptions options = new ListBlobsOptions();
+                    options.setMaxResultsPerPage(limit.intValue());
```
ikalinin1, line 291: Is there a server limitation? I think it is expected that we could receive only a limited maximum results per page

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
uarlouski: we could use simple for to stop adding elements to blobNames once it's size reaches limit

docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -169,6 +169,58 @@ When I find all blobs with name which $comparisonRule `$blobNameToMatch` in cont
 When I find all blobs with name which contains `json` in container `vividus` of storage account `testaccount` and save result to scenario variable `blobsCollection`
 ----
 
+==== Find blobs by filter
+
+Find blobs with names filtered by the specified rules in the container.
+At least one rule is required.
```
uarlouski, line 175: does it make sense to highlight it? https://docs.antora.org/antora/2.3/asciidoc/admonitions/

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
```diff
@@ -0,0 +1,82 @@
...
+    public void setResultsLimit(Optional<Integer> resultsLimit)
+    {
+        this.resultsLimit = resultsLimit;
+    }
+
+    public void validate()
+    {
+        isTrue(getResultsLimit().isPresent() || getBlobNameFilterValue().isPresent()
+                        || getBlobNameFilterRule().isPresent() || getBlobNamePrefix().isPresent(),
+                "At least one filter must be present");
```
ikalinin1, line 78:
```suggestion
        isTrue(getResultsLimit().isPresent() || getBlobNameFilterRule().isPresent() || getBlobNamePrefix().isPresent(),
                "At least one filter must be present");
```

ikalinin1: There is no reason to consider the filter value as a filter

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +238,77 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+     *                          <li><b>NEXT_BATCHES</b> - the variable will be available starting from next batch
+     *                          </ul>
+     * @param variableName      The variable name to store the list of found blob names.
+     */
+    @When("I filter blobs by:$filter in container `$containerName` of storage account `$storageAccountKey`"
+            + " and save result to $scopes variable `$variableName`")
+    public void findBlobs(BlobFilter filter, String containerName, String storageAccountKey, Set<VariableScope> scopes,
+            String variableName)
+    {
+        filter.validate();
```
ikalinin1, line 282: How do you think, maybe we could introduce the converter, it will simplify a few things:
1. Validation will be performed at the converter level
2. We could combine name filter and rule into POJO, because they work only in pairs, or we could even create the  Optional<Matcher> nameFilter right in the converter. This approach will simplify this https://github.com/vividus-framework/vividus/pull/1972/files#diff-514588a30d1bc0c47c22777fb464ba7548bd428d25bd80277f4ca8cd0981dd34R298-R302

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/BlobStorageSteps.java
```diff
@@ -231,6 +238,77 @@ public void findBlobs(StringComparisonRule rule, String blobNameToMatch, String
...
+        );
+
+        PagedIterable<BlobItem> blobItems = blobContainerClient.listBlobs(options, null);
+        Stream<String> blobNames = StreamSupport.stream(blobItems.iterableByPage().spliterator(), false)
+                .map(PagedResponse::getValue)
+                .flatMap(List::stream)
+                .map(BlobItem::getName);
+        Stream<String> filteredBlobNames = filter.getBlobNameFilterValue()
+                .map(value -> blobNames.filter(
+                        name -> filter.getBlobNameFilterRule().get().createMatcher(value).matches(name)
```
ikalinin1, line 300: I think we shouldn't create a matcher for each blob name, if there are thousands of blobs it may affect performance

ikalinin1 (approved):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/model/BlobFilter.java
tshkursk: Why user shouldn't be informed that he used step incorrectly when he specified only value.

ikalinin1: He will be notified in the second validation

uarlouski (approved):

ikalinin1 (approved):

PR #2149, opened 2021-11-05
[plugin-web-app] Delete comment attribute from proxy request body

uarlouski:
The comment attribute from [(spec)](http://www.softwareishard.com/blog/har-12-spec/#content) in HarPostData is not used anywhere

### DIFF
--- a/vividus-plugin-web-app/src/main/java/org/vividus/proxy/steps/ProxySteps.java
+++ b/vividus-plugin-web-app/src/main/java/org/vividus/proxy/steps/ProxySteps.java
@@ -288,8 +288,7 @@ public class ProxySteps
     {
         return Map.of(
                 "mimeType", postData.getMimeType(),
-                "text", postData.getText(),
-                "comment", String.valueOf(postData.getComment())
+                "text", postData.getText()
         );
     }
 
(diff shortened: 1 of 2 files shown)


ikalinin1:
Looks like breaking changes, even if this is 0,000000001 % :)

ikalinin1 (approved):

valfirst:
> Looks like breaking changes, even if this is 0,000000001 % :)

the setter for `comment` is never used in the code, so it's always `null`, variables with `null` acts in the same way as missing variables: they are not resolved

so it might look like breaking change, but it's not so

valfirst (approved):

ikalinin1:
> > Looks like breaking changes, even if this is 0,000000001 % :)
> 
> the setter for `comment` is never used in the code, so it's always `null`, variables with `null` acts in the same way as missing variables: they are not resolved
> 
> so it might look like breaking change, but it's not so

But the user is allowed to compare map-like variable against examples table and potentially he could expect null

valfirst:
> > > Looks like breaking changes, even if this is 0,000000001 % :)
> > 
> > the setter for `comment` is never used in the code, so it's always `null`, variables with `null` acts in the same way as missing variables: they are not resolved
> > 
> > so it might look like breaking change, but it's not so
> 
> But the user is allowed to compare map-like variable against examples table and potentially he could expect null

true, but not for this case, the step generates variable which contains a map with map as value, so it won't be compared against ExamplesTable or other similar data

PR #2159, opened 2021-11-08
[vividus] Fix System properties values decryption

tshkursk:


vividus/src/main/java/org/vividus/configuration/PropertiesDecryptor.java
```diff
@@ -49,6 +52,22 @@ public static Properties decryptProperties(StringEncryptor stringEncryptor, Prop
...
+    private static void processSystemProperties(Properties properties)
+    {
+        Iterator<Map.Entry<Object, Object>> iterator = properties.entrySet().iterator();
+        while (iterator.hasNext())
+        {
+            Map.Entry<Object, Object> entry = iterator.next();
+            String key = (String) entry.getKey();
+            if (key.startsWith(SYSTEM_PROPERTIES_PREFIX))
+            {
+                System.setProperty(StringUtils.removeStart(key, SYSTEM_PROPERTIES_PREFIX), (String) entry.getValue());
```
valfirst, line 68: `PropertiesDecryptor` shouldn't be responsible for setting system properties

vividus-tests/src/main/resources/properties/environment/integration/envinroment.properties
```diff
@@ -24,3 +24,5 @@ http.service-unavailable-retry.max-retries=2
 
 db.connection.csv-data.driver-class-name=org.relique.jdbc.csv.CsvDriver
 db.connection.csv-data.url=jdbc:relique:csv:classpath:data?separator=;&quoteStyle=C
+
+system.ENV_VAR=ENC(pMLhry0P0zuhIiy/lXAiiqytAAAFJSFNqAY83S9yGf8=)
```
valfirst, line 28: system properties are not the same as environment variables:
```suggestion
system.system-property=ENC(pMLhry0P0zuhIiy/lXAiiqytAAAFJSFNqAY83S9yGf8=)
```

valfirst (changes requested):

vividus/src/main/resources/org/vividus/spring-configuration.xml
```diff
@@ -3,7 +3,11 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd">
 
-    <bean id="properties" class="org.vividus.configuration.PropertiesDecryptor" factory-method="decryptProperties">
+    <bean id="properties" class="org.vividus.configuration.SystemPropertiesExtractor" factory-method="process">
```
valfirst, line 6: system properties should be set before bean factory is opened, otherwise some beans may be initialized with default system properties and race conditions are possible here

valfirst (changes requested):

ikalinin1 (approved):

vividus/src/main/java/org/vividus/configuration/BeanFactory.java
```diff
@@ -111,7 +111,7 @@ public static void close()
     {
         try
         {
-            String value = ConfigurationResolver.getInstance().getProperties()
+            String value = ConfigurationResolver.getInstance().getRawProperties()
                     .getProperty(AbstractEnvironment.ACTIVE_PROFILES_PROPERTY_NAME);
```
valfirst, line 115:
```suggestion
            String value = ConfigurationResolver.getInstance()
                    .getUndecryptedProperty(AbstractEnvironment.ACTIVE_PROFILES_PROPERTY_NAME);
```

valfirst (changes requested):

valfirst (changes requested):
It seems this change will break the ability to specify encryptor password in `overriding.properties`

valfirst (changes requested):
It seems this change will break the ability to specify encryptor password in `overriding.properties`

vividus/src/main/java/org/vividus/configuration/ConfigurationResolver.java
```diff
@@ -234,30 +233,22 @@ private static void resolveSpelExpressions(Properties properties, boolean ignore
...
+        return properties.getProperty(key);
     }
 
-    public Properties getProperties()
+    public Properties getProperties(StringEncryptor stringEncryptor)
     {
         isReset();
-        return (Properties) properties.clone();
+        PropertiesDecryptor decryptor = new PropertiesDecryptor(stringEncryptor);
+        Properties processedProperties = SystemPropertiesProcessor.process(decryptor, properties);
```
valfirst, line 250: non-encrypted system properties are set twice with this approach

valfirst (changes requested):

vividus/src/main/java/org/vividus/configuration/ConfigurationResolver.java
valfirst: the things are not changed much: the properties are not set twice, but they are still processed twice

valfirst: the proposed solution:
- find encryptor password or use a default one
- create StringEncryptor instance
- process system properties one (find by prefix, decrypt, set)

vividus/src/main/resources/org/vividus/spring-configuration.xml
```diff
@@ -26,8 +21,8 @@
     <bean id ="stringEncryptor" class="org.jasypt.encryption.pbe.StandardPBEStringEncryptor">
         <property name="config">
             <bean class="org.jasypt.encryption.pbe.config.SimplePBEConfig">
-                <property name="algorithm" value="PBEWithMD5AndDES"/>
-                <property name="password" value="#{systemProperties['vividus.encryptor.password']?:'82=thuMUH@'}"/>
+                <property name="algorithm" value="#{T(org.vividus.configuration.ConfigurationResolver).DEFAULT_ENCRYPTOR_ALGORITHM}"/>
+                <property name="password" value="#{systemProperties['#{T(org.vividus.configuration.ConfigurationResolver).VIVIDUS_ENCRYPTOR_PASSWORD_PROPERTY}']?:'#{T(org.vividus.configuration.ConfigurationResolver).DEFAULT_ENCRYPTOR_PASSWORD}'}"/>
             </bean>
         </property>
     </bean>
```
valfirst, line 28: it should be created only once (in `ConfigurationResolver`) and then it should be re-used

valfirst (changes requested):

valfirst (changes requested):

valfirst (approved):

PR #2206, opened 2021-11-16
[tests] Fix QR code step test for mobile apps

valfirst:


ikalinin1:
How does reordering fix this? Maybe something is wrong with the long-press tests?

VerkhovtsovPavel (approved):
Could you share the root cause of failure to avoid such a situation in the future, please?

uarlouski (approved):

valfirst:
1. System menu is shown on top of the application after long press tests, and the application menu cannot be found:
image
Screen Shot 2021-11-17 at 12 13 40 PM

2. Check for menu disappearing produces false-positive result on Android, because the locator catches top level header:
image
Screen Shot 2021-11-17 at 12 17 09 PM

ikalinin1 (approved):

PR #2253, opened 2021-11-29
[vividus] Add processing to incorrect 'byColumnNames' values

draker94:
If user enter incorrect column names in filtering table property 'byColumnNames' he'll get an 'IllegalArgumentException' exception.


vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -31,7 +34,13 @@
     protected List<String> filterColumnNames(List<String> allColumnNames, String byColumnNames)
     {
         List<String> filtered = new ArrayList<>(allColumnNames);
-        filtered.retainAll(Stream.of(StringUtils.split(byColumnNames, ';'))
+        String[] selectedColumnNames = StringUtils.split(byColumnNames, ';');
+
+        isTrue(allColumnNames.containsAll(Arrays.asList(selectedColumnNames)),
+                "The value of 'byColumnNames' - '%s' has one or more columns undefined in the example table",
```
valfirst, line 40: it would be good to add invalid columns to the error message

valfirst (changes requested):

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -31,7 +34,13 @@
...
+                byColumnNames);
+
+        filtered.retainAll(Stream.of(selectedColumnNames)
                 .map(String::trim)
                 .collect(Collectors.toList()));
         return filtered;
```
uarlouski, line 46: can't we just return selectedColumnNames as list?

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -31,7 +33,15 @@
     protected List<String> filterColumnNames(List<String> allColumnNames, String byColumnNames)
     {
         List<String> filtered = new ArrayList<>(allColumnNames);
-        filtered.retainAll(Stream.of(StringUtils.split(byColumnNames, ';'))
+        List<String> selectedColumnNames = Arrays.asList(StringUtils.split(byColumnNames, ';'));
```
valfirst, line 36: we need to trim column names here

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -31,7 +33,15 @@
...
+
+        String undefinedColumns = selectedColumnNames
+                .stream().filter(element -> !allColumnNames.contains(element)).collect(Collectors.joining(" "));
+        isTrue(undefinedColumns.trim().isEmpty(),
+                "The value of 'byColumnNames' has one or more columns undefined in the example table: '%s'",
```
valfirst, line 41:
```suggestion
                "'byColumnNames' refers columns missing in ExamplesTable: %s",
```

valfirst (changes requested):

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
draker94: > can't we just return selectedColumnNames as list?

I think the columns of the table should be returned in the original order, not the order that the users defined

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -31,9 +33,16 @@
     protected List<String> filterColumnNames(List<String> allColumnNames, String byColumnNames)
     {
         List<String> filtered = new ArrayList<>(allColumnNames);
```
valfirst, line 35: it makes sense to move it after the validation to avoid its creation in case validation failure

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -30,10 +32,17 @@
...
-        List<String> filtered = new ArrayList<>(allColumnNames);
-        filtered.retainAll(Stream.of(StringUtils.split(byColumnNames, ';'))
+        List<String> selectedColumnNames = Stream.of(StringUtils.split(byColumnNames, ';'))
                 .map(String::trim)
-                .collect(Collectors.toList()));
+                .collect(Collectors.toList());
+
+        String undefinedColumns = selectedColumnNames
+                .stream().filter(element -> !allColumnNames.contains(element)).collect(Collectors.joining(" "));
+        isTrue(undefinedColumns.trim().isEmpty(),
```
valfirst, line 41: so as column names are trimmed at parsing, no reason to trim here

valfirst (approved):

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -30,10 +32,17 @@
...
-        filtered.retainAll(Stream.of(StringUtils.split(byColumnNames, ';'))
+        List<String> selectedColumnNames = Stream.of(StringUtils.split(byColumnNames, ';'))
                 .map(String::trim)
-                .collect(Collectors.toList()));
+                .collect(Collectors.toList());
+
+        String undefinedColumns = selectedColumnNames
+                .stream().filter(element -> !allColumnNames.contains(element)).collect(Collectors.joining(" "));
+        isTrue(undefinedColumns.isEmpty(),
+                "'byColumnNames' refers columns missing in ExamplesTable: %s", undefinedColumns);
```
uarlouski, line 42: I think it makes sense to use line separator as separator for column names because we may want to use space as part of column name e.g.
'byColumnNames' refers columns missing in ExamplesTable:
col1
col2

uarlouski: discussed to use ; instead of space as it used less often

draker94: > discussed to use ; and space instead of space as it used less often

Yes, I think ';' + 'space symbol' will be more correct and readable

uarlouski (approved):

vividus/src/main/java/org/vividus/transformer/AbstractFilteringTableTransformer.java
```diff
@@ -30,10 +32,17 @@
...
-        List<String> filtered = new ArrayList<>(allColumnNames);
-        filtered.retainAll(Stream.of(StringUtils.split(byColumnNames, ';'))
+        List<String> selectedColumnNames = Stream.of(StringUtils.split(byColumnNames, ';'))
                 .map(String::trim)
-                .collect(Collectors.toList()));
+                .collect(Collectors.toList());
+
+        String undefinedColumns = selectedColumnNames
+                .stream().filter(element -> !allColumnNames.contains(element)).collect(Collectors.joining("; "));
+        isTrue(undefinedColumns.isEmpty(),
```
ikalinin1, line 41: Worth mentioning in documentation

docs/modules/commons/pages/table-transformers.adoc
```diff
@@ -145,6 +145,8 @@ the columns.
 
 [IMPORTANT]
 `byMaxColumns` and `byColumnNames` are competing properties and only one can be specified at the same time. The same restriction is applied to the properties `byMaxRows`, `byRandomRows` and `byRowIndexes`.
+[IMPORTANT]
+In case if `byColumnNames` contains nonexistent table columns will be thrown an exception with list of the invalid columns.
```
valfirst, line 149:
```suggestion
In case if `byColumnNames` contains nonexistent table columns the exception with list of the invalid columns will be thrown.
```

ikalinin1 (approved):

PR #2352, opened 2022-01-04
[CI] Avoid using PowerShell on Windows platform

valfirst:
Reason: PowerShell (default shell for Windows platform - https://github.blog/changelog/2019-10-17-github-actions-default-shell-on-windows-runners-is-changing-to-powershell/) doesn't fail the step in case if command is failed (this is expected default behavior - https://github.community/t/multiline-commands-on-windows-do-not-fail-if-individual-commands-fail/16753).


valfirst:
The assumptions made in this PR are wrong, the root cause is fixed in #2353

PR #2370, opened 2022-01-11
[vividus-engine] Add expression for generating null

### LINKED ISSUE #2345: introduce a new expression #{null}
introduce a new expression #{null}

_Originally posted by @ikalinin1 in https://github.com/vividus-framework/vividus/pull/2337#r776574955_

draker94:
Closes #2345

### DIFF
--- a/docs/modules/commons/pages/expressions.adoc
+++ b/docs/modules/commons/pages/expressions.adoc
@@ -419,3 +419,29 @@ When I initialize Scenario variable `listOfMaps` with values:
 |3  |
 Then `1-2-3` is = `#{evalGroovy(return listOfMaps.collect{it['key']}.sort().join('-'))}`
 ----
+
+== Null value
+
+=== `null`
+
+Represents `null` (a.k.a. `NULL`). In most case `null` means no value (see https://www.w3schools.com/sql/sql_null_values.asp[`NULL` in SQL] and https://www.json.org/json-en.html[`null` in JSON]).
+
+[source, subs="+quotes"]
+----
+#{null}
+----
+
+NOTE: Null expression can only be evaluated separately. Strings or other expressions with the nested null expression will be completely ignored and not executed.
+
+.Validation of JSON element with `null` value
+[source,gherkin]
+----
+When I initialize the scenario variable `json` with value `
+{
+    "persons": {
+        "nemo": null
+    }
+}
+`
+Then JSON element value from `${json}` by JSON path `$.persons.nemo` is equal to `#{null}`
+----
--- a/vividus-engine/src/main/java/org/vividus/steps/ParameterConvertersDecorator.java
+++ b/vividus-engine/src/main/java/org/vividus/steps/ParameterConvertersDecorator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2021 the original author or authors.
+ * Copyright 2019-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,8 @@ import org.jbehave.core.steps.ParameterConverters;
 
 public class ParameterConvertersDecorator extends ParameterConverters
 {
+    private static final String NULL_EXPRESSION = "#{null}";
+
     private final PlaceholderResolver placeholderResolver;
 
     public ParameterConvertersDecorator(Configuration configuration, PlaceholderResolver placeholderResolver)
@@ -41,6 +43,10 @@ public class ParameterConvertersDecorator extends ParameterConverters
         if (type == SubSteps.class)
         {
             return super.convert(value, type);
+        }
+        if (NULL_EXPRESSION.equals(value))
+        {
+            return null;
         }
         Object adaptedValue = placeholderResolver.resolvePlaceholders(value, type);
         if (type == String.class || adaptedValue instanceof String)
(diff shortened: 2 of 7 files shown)


docs/modules/commons/pages/expressions.adoc
```diff
@@ -419,3 +419,29 @@ When I initialize Scenario variable `listOfMaps` with values:
 |3  |
 Then `1-2-3` is = `#{evalGroovy(return listOfMaps.collect{it['key']}.sort().join('-'))}`
 ----
+
+== Null generation
+
+=== `null`
+
+Generates the `null`. Unlike the expression `eval(null)` generates not the string with `null` value, but the https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.10.7[special literal] of the null type.
```
valfirst, line 427:
```suggestion
== Null value

=== `null`

Represents `null` (a.k.a. `NULL`). In most case `null` means no value (see https://www.w3schools.com/sql/sql_null_values.asp[`NULL` in SQL] and https://www.json.org/json-en.html[`null` in JSON]).
```

docs/modules/commons/pages/expressions.adoc
```diff
@@ -419,3 +419,29 @@ When I initialize Scenario variable `listOfMaps` with values:
...
+
+[source, subs="+quotes"]
+----
+#{null}
+----
+
+NOTE: Null generation expression doesn't work in examples tables.
```
valfirst, line 434:
```suggestion
```
I guess with the proposal below it would possible to use the expression in ExamplesTable-s

docs/modules/commons/pages/expressions.adoc
```diff
@@ -419,3 +419,29 @@ When I initialize Scenario variable `listOfMaps` with values:
...
+
+.Generation `null` and comparison with JSON null
```
valfirst, line 436:
```suggestion
.Validation of JSON element with `null` value
```

docs/modules/commons/pages/expressions.adoc
```diff
@@ -419,3 +419,29 @@ When I initialize Scenario variable `listOfMaps` with values:
...
+----
+When I initialize the scenario variable `json` with value `
+{
+    "persons": {
+        "nemo": null
+    }
+}
+`
+Then JSON element value from `${json}` by JSON path `$.persons.nemo` is equal to `#{null}`
+----
```
valfirst, line 447: can we add one more example for SQL?

vividus-engine/src/main/java/org/vividus/steps/ParameterConvertersDecorator.java
```diff
@@ -43,6 +43,10 @@ public Object convert(String value, Type type)
             return super.convert(value, type);
         }
```
valfirst, line 44:
```suggestion
        }
        if (NULL_EXPRESSION.equals(value))
        {
            return null;
        }
```
?

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/steps/ExpressionAdaptor.java
```diff
@@ -74,6 +76,10 @@ public ExpressionAdaptor(StoryControls storyControls)
      */
     public Object processRawExpression(String expression)
     {
+        if (expression.equals(NULL_EXPRESSION))
```
ikalinin1, line 79: so the nested null expression is not supported? I believe there is not much of a sense in the nested #{null} expression, but since typical nested expressions are nested by design we should disclose this behaviour

valfirst: > so the nested null expression is not supported?

I agree it doesn't make much sense


> but since typical nested expressions are nested by design we should disclose this behaviour

👍 please add note to docs

docs/modules/commons/pages/expressions.adoc
draker94: > can we add one more example for SQL?

Do we really need this?
I mean, if we have next code:
`When I execute SQL query `INSERT INTO test.launch_rockets (name, payload_cost, country) VALUES ('Long March', NULL, 'China');` against `testh2db-modifiable`
When I execute SQL query `SELECT payload_cost FROM test.launch_rockets WHERE country='China'` against `testh2db-modifiable` and save result to scenario variable `null-data``
Result data will be equals this:
`Then `${null-data}` is equal to `[{payload_cost=null}]``

draker94: > can we add one more example for SQL?

> can we add one more example for SQL?

Do we really need this?
I mean, if we have next code:
`` When I execute SQL query `INSERT INTO test.launch_rockets (name, payload_cost, country) VALUES ('Long March', NULL, 'China');` against `testh2db-modifiable`
When I execute SQL query `SELECT payload_cost FROM test.launch_rockets WHERE country='China'` against `testh2db-modifiable` and save result to scenario variable `null-data` ``
Result data will be equals this:
``Then `${null-data}` is equal to `[{payload_cost=null}]` ``

valfirst:
```gherkin
When I execute SQL query `SELECT * FROM test.launch_rockets` against `vividus-db` and save result to scenario variable `launch-rockets`
Then `${launch-rockets[0].payload_cost}` is equal to `#{null}`
```
I meant something like that

valfirst: we don't need to provide a full context, just a snippet

ikalinin1 (approved):

ikalinin1 (approved):

uarlouski (approved):

PR #2397, opened 2022-01-21
[vividus-engine] Add support null values in variables 

### LINKED ISSUE #2378: Add ability to extract NULL values to variables
**Describe**
Now we can extract rows from SQL tables with NULL values:
<pre><code>When I execute SQL query `INSERT INTO test.launch_rockets (name, payload_cost, country) VALUES ('Normandy', NULL, NULL);` against `testh2db-modifiable`
When I execute SQL query `SELECT * FROM test.launch_rockets` against `testh2db-modifiable` and save result to scenario variable `test-data`
Then `${test-data}` is equal to `[{name=Normandy, payload_cost=null, country=null}]` </code></pre>
But if we want to extract the specific value of the row that contains NULL, it will not be extracted:
<pre><code>Then `${test-data[0].payload_cost}` is equal to `#{null}`</code></pre>
<pre><code>ERROR org.vividus.softassert.SoftAssert - Fail: Potentially known issue: VVD-6 (Type: INTERNAL. Status: UNKNOWN). Checking if "${test-data[0].payload_cost}" is equal to "null" [Expected: "null" Actual: was "${test-data[0].payload_cost}"]</code></pre>
Non-null values are retrieved normally:
<pre><code>Then `${test-data[0].name}` is equal to `Normandy`</code></pre>
<pre><code>INFO  org.vividus.softassert.SoftAssert - Pass: Checking if "Normandy" is equal to "Normandy" [Expected: "Normandy" Actual: was "Normandy"]</code></pre>

Add ability to extract NULL values from SQL tables to variables.

_Originally posted by @valfirst in https://github.com/vividus-framework/vividus/pull/2370#discussion_r784122560_

draker94:
Closes #2378
(Only logic)


vividus-engine/src/main/java/org/vividus/steps/VariableResolver.java
```diff
@@ -147,6 +148,10 @@ private Object replaceVariableKeyWithValue(String target, String variableKey)
         {
             return target;
         }
+        if (Optional.of(NULL_VALUE).equals(variableValue))
```
valfirst, line 151: this constant is a kind of "internal flag", but there is a chance variable will contain `nullVariable` value and this logic will fail

draker94: > this constant is a kind of "internal flag", but there is a chance variable will contain `nullVariable` value and this logic will fail
In this case in VariableResolver.java we will get exactly `nullValue `but not `Optional.of("nullValue")`

draker94: > this constant is a kind of "internal flag", but there is a chance variable will contain `nullVariable` value and this logic will fail

In this case in VariableResolver.java we will get exactly `nullValue `but not `Optional.of("nullValue")`

vividus-engine/src/main/java/org/vividus/steps/ParameterConvertersDecorator.java
```diff
@@ -49,6 +49,10 @@ public Object convert(String value, Type type)
             return null;
         }
         Object adaptedValue = placeholderResolver.resolvePlaceholders(value, type);
+        if (adaptedValue == null)
+        {
+            return null;
+        }
```
valfirst, line 55: taking into account recent commits, this change is not needed anymore

vividus-engine/src/main/java/org/vividus/steps/VariableResolver.java
```diff
@@ -142,11 +142,20 @@ private Object resolveVariable(String value, String variableKey)
...
     private Object replaceVariableKeyWithValue(String target, String variableKey)
     {
-        Object variableValue = getVariableValue(variableKey);
-        if (variableValue == null)
+        Object variableValue;
+        try
         {
+            variableValue = getVariableValue(variableKey);
+        }
+        catch (IllegalArgumentException e)
```
valfirst, line 150: we should never catch runtime exceptions (especially very generic `IllegalArgumentException`)

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/steps/ParameterConvertersDecorator.java
```diff
@@ -44,11 +44,11 @@ public Object convert(String value, Type type)
         {
             return super.convert(value, type);
         }
-        if (NULL_EXPRESSION.equals(value))
+        Object adaptedValue = placeholderResolver.resolvePlaceholders(value, type);
+        if (NULL_EXPRESSION.equals(adaptedValue))
         {
             return null;
         }
```
valfirst, line 51: move this logic to `PlaceholderResolver`

vividus-engine/src/main/java/org/vividus/variable/Variables.java
```diff
@@ -36,6 +36,7 @@
     private static final int VARIABLE_NAME_GROUP = 1;
     private static final int LIST_INDEX_GROUP = 2;
     private static final int MAP_KEY_GROUP = 3;
+    private static final String VIVIDUS_NULL_EXPRESSION = "#{null}";
```
valfirst, line 39: make `public` and re-use in `PlaceholderResolver` (after moving the logic)

valfirst (changes requested):

valfirst (approved):

PR #2407, opened 2022-01-24
[vividus] Optimize stories search for JUnit runner

uarlouski:


vividus/src/main/java/org/vividus/runner/GenericRunner.java
```diff
@@ -51,6 +52,8 @@
     private final Embedder embedder;
     private final IBatchedPathFinder batchedPathFinder;
 
+    private Map<String, List<String>> batchToStoryPaths;
```
valfirst, line 55: this logic should be moved to `BatchedPathFinder`

valfirst (changes requested):

vividus/src/main/java/org/vividus/runner/GenericRunner.java
uarlouski: Why?  It still won't be clear from the main logic that this value gets cached, so each time someone gets to this point he will have to lookup BatchedPathFinder class

vividus-engine/src/main/java/org/vividus/BatchedPathFinder.java
```diff
@@ -25,25 +25,35 @@
 import org.vividus.batch.BatchResourceConfiguration;
 import org.vividus.batch.BatchStorage;
 
+/**
+ * The API provided by this class <b><u>is not</u></b> thread-safe and expected to be invoked from the main thread
```
valfirst, line 29:
```suggestion
 * The API provided by this class is <b><u>not</u></b> thread-safe and expected to be invoked from the main thread
```

valfirst (changes requested):

valfirst (approved):

vividus-engine/src/test/java/org/vividus/BatchedPathFinderTests.java
```diff
@@ -54,8 +50,9 @@ void testFindPaths() throws IOException
         when(batchStorage.getBatchResourceConfigurations()).thenReturn(Map.of(batchKey, batchResourceConfiguration));
         List<String> testPaths = List.of("testPath");
         when(pathFinder.findPaths(batchResourceConfiguration)).thenReturn(testPaths);
-        Map<String, List<String>> actual = batchedPathFinder.findPaths();
         Map<String, List<String>> expected = Map.of(batchKey, testPaths);
-        assertEquals(expected, actual);
+        assertEquals(expected, batchedPathFinder.getPaths());
```
ikalinin1, line 54: how caching is tested? Shouldn't you verify invocation quantity?

uarlouski: please check verifyNoMoreInteractions below

ikalinin1 (approved):

valfirst (approved):

Issue #438, opened 2020-03-27
Implement JBehave-JUnit-Runner for Vividus


valfirst:
Current `jbehave-junit-runner` is strongly relies on jbehave-core and can't take into account Vividus features like batches, per-batch configuration of meta-filters and etc.

Possible solutions:
- fork and extend `jbehave-junit-runner`
- implement own runner

dKatechev:
Scenarios with Examples table are not running properly with Eclipse JUnit 4 test runner.
Test execution is failing with the exception below:

`2020-04-21 12:30:53,481 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - Failed to run story file:/D:/crepo/commerce-api/v-seph/instagram-tests/bin/main/story/seph/instagram/management/CE-301 Management Service - Accept Order Record.story
2020-04-21 12:30:53,482 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - java.lang.NullPointerException
2020-04-21 12:30:53,483 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - at com.github.valfirst.jbehave.junit.monitoring.JUnitScenarioReporter$TestState.getNextOrNull(JUnitScenarioReporter.java:408)
2020-04-21 12:30:53,483 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - at com.github.valfirst.jbehave.junit.monitoring.JUnitScenarioReporter$TestState.moveToNextExample(JUnitScenarioReporter.java:392)
2020-04-21 12:30:53,483 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - at com.github.valfirst.jbehave.junit.monitoring.JUnitScenarioReporter$TestState.access$1300(JUnitScenarioReporter.java:368)
2020-04-21 12:30:53,483 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - at com.github.valfirst.jbehave.junit.monitoring.JUnitScenarioReporter.example(JUnitScenarioReporter.java:234)
2020-04-21 12:30:53,483 [batch-1-thread-1] INFO  org.vividus.bdd.log.LoggingEmbedderMonitor - at org.jbehave.core.reporters.DelegatingStoryReporter.example(DelegatingStoryReporter.java:143)
[...]

PR #2420, opened 2022-01-27
[vividus] Decouple generic runner from batch mechanism

uarlouski:


vividus/src/main/java/org/vividus/runner/AbstractGenericRunner.java
```diff
@@ -33,97 +27,61 @@
...
-import org.vividus.BatchedEmbedder;
-import org.vividus.IBatchedPathFinder;
 import org.vividus.IRunStatusProvider;
 import org.vividus.Status;
 import org.vividus.configuration.BeanFactory;
 import org.vividus.configuration.Vividus;
 import org.vividus.log.TestInfoLogger;
 
-public class GenericRunner extends JUnitStories
+public abstract class AbstractGenericRunner extends JUnitStories
```
ikalinin1, line 36: `AbstractGenericRunner` -> `AbstractRunner`

vividus/src/main/java/org/vividus/runner/AbstractGenericRunner.java
```diff
@@ -33,97 +27,61 @@
...
 {
     private static final int RUN_PASSED_EXIT_CODE = 0;
     private static final int RUN_KNOWN_ISSUES_EXIT_CODE = 1;
     private static final int RUN_FAILED_EXIT_CODE = 2;
     private static final int ERROR_EXIT_CODE = 3;
 
-    private static String embedderBeanName;
-    private final Embedder embedder;
-    private final IBatchedPathFinder batchedPathFinder;
+    private static Class<?> runnerClass;
```
ikalinin1, line 43: final?

vividus/src/main/java/org/vividus/runner/AbstractGenericRunner.java
```diff
@@ -33,97 +27,61 @@
...
-        try
-        {
-            return batchedPathFinder.getPaths();
-        }
-        catch (IOException e)
-        {
-            throw new UncheckedIOException(e);
-        }
-    }
+    protected abstract Embedder getEmbedder();
```
valfirst, line 75: just use `configuredEmbedder()`

vividus/src/test/java/org/vividus/runner/GenericRunnerTests.java
```diff
@@ -35,24 +35,41 @@
     @Test
```
valfirst, line 35: the name of the test class should be updated properly

valfirst (changes requested):

vividus/src/main/java/org/vividus/runner/AbstractGenericRunner.java
uarlouski: okay, but I added getEmbedder because in the parent class configuredEmbedder() is not required to override, so we may forget it in the future

vividus/src/main/java/org/vividus/runner/AbstractRunner.java
```diff
@@ -16,114 +16,67 @@
...
+        return configuredEmbedder().stepsFactory();
     }
 
-    public static void main(String[] args)
+    @SuppressWarnings("NoMainMethodInAbstractClass")
+    public static void main(String[] args) throws InitializationError
     {
-        Class<?> clazz = MethodHandles.lookup().lookupClass();
-        Result result = new JUnitCore().run(clazz);
+        Runner runner = new JUnit4(runnerClass);
```
valfirst, line 74: please describe why this change is needed

uarlouski: Let me split into several parts:
1) There is a bug in our mechanic that shows up when we run tests using JUnit4StoryRunner and when we have two stories from different batches where the first batch inits NEXT_BATCHES variable (var) and the second batch tries to load table from the saved variable, in this case variable is not resolved when building JUnitTree and as a result we get
```
java.lang.IllegalArgumentException: java.io.FileNotFoundException: class path resource [${var}] cannot be opened because it does not exist
	at org.vividus.resource.StoryLoader.resourceAsStream(StoryLoader.java:63)
	at org.jbehave.core.io.LoadFromClasspath.loadResourceAsText(LoadFromClasspath.java:67)
	at org.vividus.resource.StoryLoader.loadResourceAsText(StoryLoader.java:51)
	at org.jbehave.core.model.ExamplesTableFactory.createExamplesTable(ExamplesTableFactory.java:84)
	at org.jbehave.core.parsers.RegexStoryParser.parseExamplesTable(RegexStoryParser.java:332)
	at org.jbehave.core.parsers.RegexStoryParser.parseScenario(RegexStoryParser.java:292)
	at org.jbehave.core.parsers.RegexStoryParser.parseScenariosFrom(RegexStoryParser.java:269)
	at org.jbehave.core.parsers.RegexStoryParser.parseStory(RegexStoryParser.java:69)
	at org.jbehave.core.embedder.PerformableTree.storyOfPath(PerformableTree.java:329)
	at org.jbehave.core.embedder.StoryManager.storyOfPath(StoryManager.java:65)
	at org.jbehave.core.embedder.StoryManager.storiesOfPaths(StoryManager.java:71)
	at org.jbehave.core.junit.JUnit4StoryRunner.createPerformableTree(JUnit4StoryRunner.java:136)
	at org.jbehave.core.junit.JUnit4StoryRunner.buildDescriptionsFromStories(JUnit4StoryRunner.java:121)
	at org.jbehave.core.junit.JUnit4StoryRunner.<init>(JUnit4StoryRunner.java:44)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.junit.internal.builders.AnnotatedBuilder.buildRunner(AnnotatedBuilder.java:104)
	at org.junit.internal.builders.AnnotatedBuilder.runnerForClass(AnnotatedBuilder.java:86)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:70)
	at org.junit.internal.builders.AllDefaultPossibilitiesBuilder.runnerForClass(AllDefaultPossibilitiesBuilder.java:37)
	at org.junit.runners.model.RunnerBuilder.safeRunnerForClass(RunnerBuilder.java:70)
	at org.junit.internal.requests.ClassRequest.createRunner(ClassRequest.java:28)
	at org.junit.internal.requests.MemoizingRequest.getRunner(MemoizingRequest.java:19)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.createUnfilteredTest(JUnit4TestLoader.java:90)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.createTest(JUnit4TestLoader.java:76)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestLoader.loadTests(JUnit4TestLoader.java:49)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:513)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:756)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:452)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)
Caused by: java.io.FileNotFoundException: class path resource [${var}] cannot be opened because it does not exist
	at org.springframework.core.io.ClassPathResource.getInputStream(ClassPathResource.java:199)
	at org.vividus.resource.StoryLoader.resourceAsStream(StoryLoader.java:59)
	... 31 more
```
This issue only reproduces when JUnit4StoryRunner is used
2) when we run tests using gradle task the default runner is Junit4 since the GenericRunner class is not annotated with RunWith annotation, so the above but is not reproduced since we using different runners
3) once we move common logic into StoriesRunner we also should use this as a test class in JUnitCore().run(clazz);, but since this class is annotated with @RunWith annotation, the default runner becomes JUnit4StoryRunner that is redundant when running from command line since we don't need JunitTree and additionaly we get the error in the first case.

uarlouski: related to https://github.com/vividus-framework/vividus/issues/438

ikalinin1 (approved):

valfirst (approved):

PR #2421, opened 2022-01-28
Restrict TestResourceLoader from search of non-classpath resources in subfolders

skyoker:


vividus-engine/src/main/java/org/vividus/resource/TestResourceLoader.java
```diff
@@ -54,9 +54,18 @@ public TestResourceLoader(IPropertyParser propertyParser, ResourcePatternResolve
...
+            Resource[] allResources = null;
+            if (resourceLocation.startsWith(FILE_URL_PREFIX))
+            {
+                locationPattern = normalizedResourceLocation;
+                allResources = resourcePatternResolver.getResources(locationPattern + resourcePattern);
+            }
+            else
+            {
+                locationPattern = CLASSPATH_ALL_URL_PREFIX + normalizedResourceLocation;
+                allResources = resourcePatternResolver.getResources(locationPattern + "**/" + resourcePattern);
```
uarlouski, line 67: @valfirst as I see in default behavior we add **/, but should not the user be responsible for adding this into resource pattern on his own?

valfirst: tricky question, but "backward compatibility" is our goal

vividus-engine/src/main/java/org/vividus/resource/TestResourceLoader.java
```diff
@@ -54,9 +54,18 @@ public TestResourceLoader(IPropertyParser propertyParser, ResourcePatternResolve
...
+            }
```
valfirst, line 68:
```suggestion
            String locationPattern;
            String fullLocationPattern;            
            if (resourceLocation.startsWith(FILE_URL_PREFIX))
            {
                locationPattern = normalizedResourceLocation;
                fullLocationPattern = locationPattern + resourcePattern;
            }
            else
            {
                locationPattern = CLASSPATH_ALL_URL_PREFIX + normalizedResourceLocation;
                fullLocationPattern = locationPattern + "**/" + resourcePattern;
            }
            Resource[] allResources = resourcePatternResolver.getResources(fullLocationPattern);
```

ikalinin1 (approved):

valfirst (approved):

uarlouski (approved):

PR #2424, opened 2022-01-28
[agent-reportportal] Do not report comments

ikalinin1:
Reporting of the comments in any of the available statuses will affect overall statistics count.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -58,4 +58,3 @@ include 'vividus-exporter-commons'
 
 include 'vividus-docker-bundler'
 include 'vividus-tests'
-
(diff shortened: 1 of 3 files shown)


settings.gradle
```diff
@@ -58,4 +58,3 @@ include 'vividus-exporter-commons'
 
 include 'vividus-docker-bundler'
 include 'vividus-tests'
-
```
valfirst, line 61: is it expected?

valfirst (changes requested):

settings.gradle
uarlouski: this change bothers me locally as well

ikalinin1: spotless does it

ikalinin1: should... abide.. spotlesssss

uarlouski (approved):

PR #2521, opened 2022-02-23
[plugin-applitools] Do not publish Applitools HTTP requests to allure report

uarlouski:


vividus-http-client/src/main/java/org/vividus/http/client/LoggingHttpRequestInterceptor.java
```diff
@@ -0,0 +1,26 @@
...
+import org.apache.http.HttpRequestInterceptor;
+import org.apache.http.protocol.HttpContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingHttpRequestInterceptor implements HttpRequestInterceptor
+{
+    private final Logger logger;
+
+    public LoggingHttpRequestInterceptor(String loggerName)
```
valfirst, line 16: it makes sense to add constructor with `Class` as well

vividus-plugin-applitools/src/main/resources/spring.xml
```diff
@@ -42,6 +42,11 @@
                 <property name="lastRequestInterceptor"><null/></property>
                 <property name="lastResponseInterceptor"><null/></property>
                 <property name="httpResponseHandlers"><list/></property>
+                <property name="firstRequestInterceptor">
+                    <bean class="org.vividus.http.client.LoggingHttpRequestInterceptor">
+                        <constructor-arg value="org.vividus.visual.eyes.ApplitoolsLogging" />
```
valfirst, line 47:
```suggestion
                        <constructor-arg value="org.vividus.visual.eyes.ApplitoolsHttpClient" />
```

vividus/src/main/resources/log4j2.xml
```diff
@@ -39,6 +39,10 @@
             <AppenderRef ref="console" />
             <AppenderRef ref="file" />
         </Logger>
+        <Logger name="org.vividus.visual.eyes.ApplitoolsLogging" level="INFO" additivity="false">
```
valfirst, line 42: it should be configured in the plugin

valfirst (changes requested):

vividus-http-client/src/main/java/org/vividus/http/client/HttpClient.java
```diff
@@ -82,7 +78,6 @@ public HttpResponse execute(HttpUriRequest request) throws IOException
     @Override
     public HttpResponse execute(HttpUriRequest request, HttpContext context) throws IOException
     {
-        LOGGER.info("{}", request);
```
ikalinin1, line 85: Why the same approach as skipResponseEntity doesn't work in this case?
What should I do if I want to enable console logging in the other place?

uarlouski: - I think because skip response entity directly affects the response that we return, whereas logging doesn't.
- I also have doubts about introducing new boolean switch because we "add" one more logger disabling mechanism in addition to one provided by logging itself
- I think you meant allure logging, you will have to configure one more logger and register new http client bean
- anyway I'll consider this since your suggestion seems more concise

uarlouski: Introducing of new property won't help since we still the logging to go into console and file appenders, so it requires configuring a new logger

ikalinin1: Yeah, in simple words can we design it to be applicable by introducing changes in the single place, not the two xmls

uarlouski: why two xmls? once we have such real case we will definitely try to reuse it :)

vividus-http-client/src/main/java/org/vividus/http/client/LoggingHttpRequestInterceptor.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingHttpRequestInterceptor implements HttpRequestInterceptor
+{
+    private final Logger logger;
+
+    public LoggingHttpRequestInterceptor(Class<?> loggerType)
+    {
+        this(loggerType.getName());
```
valfirst, line 34: this.logger = LoggerFactory.getLogger(loggerType);

vividus-plugin-applitools/src/main/resources/spring.xml
```diff
@@ -42,6 +42,11 @@
                 <property name="lastRequestInterceptor"><null/></property>
                 <property name="lastResponseInterceptor"><null/></property>
                 <property name="httpResponseHandlers"><list/></property>
+                <property name="firstRequestInterceptor">
+                    <bean class="org.vividus.http.client.LoggingHttpRequestInterceptor">
+                        <constructor-arg value="org.vividus.visual.eyes.ApplitoolsHttpClient" />
+                    </bean>
+                </property>
```
valfirst, line 49:
```suggestion
                <property name="firstRequestInterceptor">
                    <bean class="org.vividus.http.client.LoggingHttpRequestInterceptor">
                        <constructor-arg value="org.vividus.visual.eyes.ApplitoolsHttpClient" />
                    </bean>
                </property>
                <property name="lastRequestInterceptor"><null/></property>
                <property name="lastResponseInterceptor"><null/></property>
                <property name="httpResponseHandlers"><list/></property>

```

vividus-plugin-applitools/src/main/resources/spring.xml
```diff
@@ -42,6 +42,11 @@
                 <property name="lastRequestInterceptor"><null/></property>
                 <property name="lastResponseInterceptor"><null/></property>
                 <property name="httpResponseHandlers"><list/></property>
+                <property name="firstRequestInterceptor">
+                    <bean class="org.vividus.http.client.LoggingHttpRequestInterceptor">
+                        <constructor-arg value="org.vividus.visual.eyes.ApplitoolsHttpClient" />
```
valfirst, line 47: no need to specify constructor arg type?

valfirst (changes requested):

vividus-plugin-applitools/src/main/resources/spring.xml
uarlouski: since we use value the String constructor will be used

valfirst (approved):

ikalinin1 (approved):

PR #2542, opened 2022-03-03
Fix GenericNestedSteps to not execute steps if no elements found

skyoker:


vividus-extension-selenium/src/main/java/org/vividus/steps/ui/GenericNestedSteps.java
```diff
@@ -114,12 +114,18 @@ public void performAllStepsWhileElementsExist(ComparisonRule comparisonRule, int
...
             firstIteration.setValue(false);
-            return softAssert.assertThat("Elements number", getElementsNumber(locator), elementsMatcher);
+            elementsMatch = softAssert.assertThat("Elements number", foundElements, elementsMatcher);
         }
-        return elementsMatcher.matches(getElementsNumber(locator));
+        else
+        {
+            elementsMatch = elementsMatcher.matches(foundElements);
+        }
+        return elementsMatch && foundElements != 0;
```
valfirst, line 128: what if step conditions is `= 0`? I guess we should either add the validation of the condition or change the step syntax

valfirst (changes requested):

PR #2556, opened 2022-03-07
Use latest Azure REST API version for all resource management steps

valfirst:


vividus-plugin-azure-event-grid/src/main/java/org/vividus/azure/eventgrid/EventGridManagementSteps.java
```diff
@@ -16,40 +16,36 @@
...
-    public EventGridManagementSteps(AzureProfile azureProfile, TokenCredential tokenCredential,
-            InnersJacksonAdapter innersJacksonAdapter, VariableContext variableContext)
+    public EventGridManagementSteps(AzureProfile azureProfile, TokenCredential tokenCredential, SoftAssert softAssert,
+            VariableContext variableContext)
     {
-        this.eventGridManager = EventGridManager.authenticate(tokenCredential, azureProfile);
-        this.variableContext = variableContext;
-        this.innersJacksonAdapter = innersJacksonAdapter;
+        super(HttpPipelineProvider.buildHttpPipeline(tokenCredential, azureProfile),
+                azureProfile.getEnvironment().getResourceManagerEndpoint(), "2021-12-01", softAssert,
```
uarlouski, line 39: Shouldn't we let end users to change the api version as per their needs?

uarlouski (approved):

ikalinin1 (approved):

vividus-extension-azure/src/main/java/org/vividus/azure/resourcemanager/AbstractAzureResourceManagementSteps.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    protected void saveHttpResponseAsVariable(String urlPath, Set<VariableScope> scopes, String variableName)
+    {
+        // Workaround for https://github.com/Azure/azure-sdk-for-java/issues/27268
+        String url = resourceManagerEndpoint + urlPath + "?api-version=" + apiVersion;
+        HttpRequest httpRequest = new HttpRequest(HttpMethod.GET, url);
+        try (HttpResponse httpResponse = httpPipeline.send(httpRequest).block())
+        {
+            String responseBody = httpResponse.getBodyAsString().block();
```
ikalinin1, line 57: SQ says NPE is possible here, it is false negative?

valfirst: yes, it's false-positive, the reason is @Nullable annotation, but it won't happen with the current SDK implementation

PR #2560, opened 2022-03-10
[vividus] Clarify execution plan log message

valfirst:


vividus/src/main/java/org/vividus/log/TestInfoLogger.java
```diff
@@ -101,7 +101,7 @@ public static void logExecutionPlan(Map<String, List<String>> executionPlan)
             try (Formatter message = new Formatter())
             {
                 message.format(separatorFormat, HORIZONTAL_RULE);
-                message.format("%n Execution plan:");
+                message.format("%n Execution plan (before filtering by meta):");
```
ikalinin1, line 104: Can't we filter them out? The meta is static, right?

ikalinin1 (approved):

vividus/src/main/java/org/vividus/log/TestInfoLogger.java
valfirst: Tricky question:
1) we can filter at both story and scenario levels
2) filtering is done during test execution
3) we log filtered out stories as separate entries

uarlouski (approved):

PR #2564, opened 2022-03-11
[vividus] Publish unified diff for strings comparison

### LINKED ISSUE #2584: Add possibility to compare text files and see their differences if there is any
As a user I want to be able to compare two different text files with the possibility to see concrete places of differences if there is any
Example of use:
I download blob from Azure Storage Account
I make a text file via template and certain data
I compare these two files to verify they are identical and if it isn't so, I want to see concrete items of difference

ikalinin1:
Closes #2584
Screenshot 2022-03-11 at 15 28 53


valfirst (changes requested):

vividus/src/main/java/org/vividus/steps/VariablesSteps.java
```diff
@@ -60,7 +61,12 @@ public VariablesSteps(VariableContext variableContext, ISoftAssert softAssert,
             {
                 String readableCondition = EnumUtils.toHumanReadableForm(condition);
                 String description = "Checking if \"" + value1 + "\" is " + readableCondition + " \"" + value2 + "\"";
-                return softAssert.assertThat(description, value1, condition.getComparisonRule(value2));
+                boolean passed = softAssert.assertThat(description, value1, condition.getComparisonRule(value2));
+                if (!passed)
+                {
+                    diffAttachmentPublisher.publishDiff(value1, value2);
```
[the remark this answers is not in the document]
ikalinin1: I can't agree here, it might be useful in the case of invisible chars and other cases, because it highlights corner cases much better then plain old assertion

docs/modules/ROOT/pages/tests-configuration.adoc
```diff
@@ -81,6 +81,10 @@ NOTE: The properties marked with *bold* are mandatory.
 |`statistics.print-failures`
 |`false`
 |If set to `true` prints table of failures including: story, scenario, step and error message
+
+|`report.text-length-diff-threshold`
+|`100`
+|Defines text threshold length to enable unified diff report publishing on failure, for variable comparisons
```
valfirst, line 87: I think it makes sense to move this description to the relevant steps to make it transparent for users.

valfirst (changes requested):

docs/modules/ROOT/pages/tests-configuration.adoc
ikalinin1: We don't have a docs for the step, isn't it out of scope this PR ? :)

vividus/src/main/java/org/vividus/publishing/DiffAttachmentPublisher.java
```diff
@@ -0,0 +1,78 @@
...
+    public DiffAttachmentPublisher(AttachmentPublisher attachmentPublisher)
+    {
+        this.attachmentPublisher = attachmentPublisher;
+    }
+
+    public void publishDiff(Object expected, Object actual)
+    {
+        String expectedString = String.valueOf(expected);
+        String actualString = String.valueOf(actual);
+        if (expectedString.length() < textLengthDiffThreshold && actualString.length() < textLengthDiffThreshold)
```
valfirst, line 49: one of the code branches is not covered by unit tests

valfirst (approved):

uarlouski (approved):

PR #2664, opened 2022-04-14
[vividus] Move status code calculation into results provider

uarlouski:


vividus/src/main/java/org/vividus/results/ResultsProvider.java
```diff
@@ -20,14 +20,13 @@
...
 
-import org.vividus.report.allure.model.Status;
 import org.vividus.results.model.Failure;
 import org.vividus.results.model.NodeType;
 import org.vividus.results.model.Statistic;
 
 public interface ResultsProvider
 {
-    Optional<Status> getRunStatus();
+    int getStatusCode();
```
valfirst, line 29:
```suggestion
    int calculateExitCode();
```

vividus/src/main/java/org/vividus/runner/AbstractTestRunner.java
```diff
@@ -32,14 +31,10 @@
...
 import org.vividus.log.TestInfoLogger;
-import org.vividus.report.allure.model.Status;
 import org.vividus.results.ResultsProvider;
 
 public abstract class AbstractTestRunner extends JUnitStories
 {
-    private static final int RUN_PASSED_EXIT_CODE = 0;
-    private static final int RUN_KNOWN_ISSUES_EXIT_CODE = 1;
-    private static final int RUN_FAILED_EXIT_CODE = 2;
     private static final int ERROR_EXIT_CODE = 3;
```
valfirst, line 38: create an enum for all exit codes

valfirst (changes requested):

vividus/src/main/java/org/vividus/results/CollectingStatisticsStoryReporter.java
```diff
@@ -104,9 +108,19 @@ public Optional<List<Failure>> getFailures()
...
+        if (runStatus.isEmpty())
+        {
+            return RUN_FAILED_EXIT_CODE;
+        }
+        switch (runStatus.get())
+        {
+            case PASSED:            return RUN_PASSED_EXIT_CODE;
+            case KNOWN_ISSUES_ONLY: return RUN_KNOWN_ISSUES_EXIT_CODE;
+            default:                return RUN_FAILED_EXIT_CODE;
+        }
```
ikalinin1, line 123:
```suggestion
       return status.get().map(s -> {
          switch (s)
          {
              case PASSED:            return RUN_PASSED_EXIT_CODE;
              case KNOWN_ISSUES_ONLY: return RUN_KNOWN_ISSUES_EXIT_CODE;
              default:                return RUN_FAILED_EXIT_CODE;
          }
        }).orElse(RUN_FAILED_EXIT_CODE)
     
```

vividus/src/main/java/org/vividus/results/ResultsProvider.java
uarlouski: this is exit code in the context of System.exit(...) but as per docs it also called status code that less coupled to exit op

vividus/src/main/java/org/vividus/results/CollectingStatisticsStoryReporter.java
uarlouski: prefer initial variant as the switch in map doesn't look good (just my opinion)

valfirst (changes requested):

valfirst (approved):

ikalinin1 (approved):

PR #2681, opened 2022-04-19
Make allure plugin loading more flexible

skyoker:
This redesign is needed to have convenient way of adding custom non-vividus allure plugins.


vividus-allure-adaptor/src/main/resources/org/vividus/report/allure/spring.xml
```diff
@@ -1,13 +1,20 @@
...
 <beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="http://www.springframework.org/schema/beans"
-    xsi:schemaLocation="http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd"
+    xmlns:context="http://www.springframework.org/schema/context"
+    xsi:schemaLocation="
+     http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd
+     http://www.springframework.org/schema/context https://www.springframework.org/schema/context/spring-context.xsd"
     default-lazy-init="true">
 
+    <context:component-scan base-package="org.vividus.report.allure.plugin" />
```
valfirst, line 10: please, declare all plugins beans explicitly instead of using `component-scan`

vividus-allure-adaptor/src/test/java/org/vividus/report/allure/adapter/AllureReportGeneratorTests.java
```diff
@@ -216,6 +218,7 @@ private LoggingEvent buildReportGeneratedLogEvent(Path reportDirectory)
     private void testEnd(File reportDirectory) throws IOException
     {
         Path historyDirectory = tempDir.resolve("history");
+        List<Plugin> plugins = List.of();
```
valfirst, line 221: it's better to add some plugin mock to this list

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/AllurePluginsProvider.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.vividus.report.allure;
+
+import java.util.List;
+
+import javax.inject.Inject;
+
+import io.qameta.allure.core.Plugin;
+
+public class AllurePluginsProvider
```
valfirst, line 25: do we need this list container? can't plugins list be injected into `AllureReportGenerator` constructor?

vividus-allure-adaptor/src/main/resources/org/vividus/report/allure/spring.xml
valfirst: just declare bean:
```xml
<bean class="my.allure.Plugin" />
```
that's it

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/plugin/VividusLogoPlugin.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.vividus.report.allure.plugin;
+
+import java.util.List;
+
+import javax.inject.Named;
+
+@Named
+public class VividusLogoPlugin extends EmbeddedPlugin
```
ikalinin1, line 24: Any benefits of having a class? Why not put this into the xml config?

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/AllurePluginsProvider.java
skyoker: This provider is needed to have ability to manage allure plugins on runtime.

valfirst: could you please elaborate on this?

vividus-allure-adaptor/src/test/java/org/vividus/report/allure/adapter/AllureReportGeneratorTests.java
skyoker: The problem here is that tests for AllureReportGenerator are not atomic and this list will be used inside external allure logic. So with mocking of list we will need additionally mock all interactions with this mock.

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/plugin/VividusLogoPlugin.java
skyoker: The only benefit I see (from my point of view) is that it's easier to configure plugins via java code. Potentially it will result in more simple support:)

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/AllurePluginsProvider.java
skyoker: The problem here is that without direct defining of list of beans spring generates different instances of lists for all places where this beans should be used. As the result you will not have access to list instance in AllureReportGenerator, so there will not be a possibility to dynamically change this list.

valfirst: ok, that's clear, but why do you need to change the list dynamically?

vividus-allure-adaptor/src/main/java/org/vividus/report/allure/plugin/VividusLogoPlugin.java
ikalinin1: Can you please declare it in the config then? 
Because I believe "the best code is no code at all." :)

vividus-allure-adaptor/src/test/java/org/vividus/report/allure/plugin/EmbeddedPluginTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+import io.qameta.allure.Extension;
+
+@ExtendWith(MockitoExtension.class)
+class EmbeddedPluginTests
+{
+    private static final String ID = "id";
+
+    @Mock
+    private List<String> resourceList;
```
uarlouski, line 36: do we really need to mock collection here?

vividus-allure-adaptor/src/main/resources/org/vividus/report/allure/spring.xml
```diff
@@ -1,13 +1,47 @@
...
-    xmlns="http://www.springframework.org/schema/beans"
-    xsi:schemaLocation="http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd"
+    xmlns="http://www.springframework.org/schema/beans" xmlns:util="http://www.springframework.org/schema/util"
+    xsi:schemaLocation="
+     http://www.springframework.org/schema/beans https://www.springframework.org/schema/beans/spring-beans.xsd
+     http://www.springframework.org/schema/util https://www.springframework.org/schema/util/spring-util.xsd"
     default-lazy-init="true">
 
+    <bean class="org.vividus.report.allure.plugin.EmbeddedPlugin">
+        <constructor-arg><value>behaviors</value></constructor-arg>
```
uarlouski, line 10: why not <constructor-arg value="behaviors" />?

vividus-allure-adaptor/src/test/java/org/vividus/report/allure/plugin/EmbeddedPluginTests.java
skyoker: Why not? What is the alternative?

uarlouski: I think it may cause side effects since there are too many methods (List + Collection AP) that allure may rely on and if it doesn't fail now it  may fail in the future, old post on side effect example https://stackoverflow.com/questions/18483176/mockito-mocking-an-arraylist-that-will-be-looped-in-a-for-loop
Why not to pass empty list?

ikalinin1 (approved):

uarlouski (approved):

valfirst (approved):

PR #2686, opened 2022-04-21
[plugin-azure-resource-manager] Add steps to create/update/delete Azure resources

valfirst:


docs/modules/plugins/pages/plugin-azure-resource-manager.adoc
```diff
@@ -69,3 +70,80 @@ resourceGroups/sample-resource-group/providers/Microsoft.KeyVault/vaults/sample-
 ----
 When I get Azure resource with identifier `resourceGroups/TEST-RG/providers/Microsoft.KeyVault/vaults/KEY-VAULT-NAME` using API version `2021-10-01` and save it to scenario variable `key-vault-properties`
 ----
+
+==== Create/Update Azure resource
```
ikalinin1, line 74: If the steps are used without necessary permissions what kind of error will the output be in that case? Will it be transparent for the user?

valfirst: I guess the permissions are needed, but they are either global and managed via roles, or very specific and granted per resource

ikalinin1 (approved):

uarlouski (approved):

PR #2692, opened 2022-04-22
Standardize table transformers API

valfirst:


docs/modules/ROOT/pages/glossary.adoc
```diff
@@ -47,3 +47,41 @@ can be referenced in the tests as `/data/request.json`.
 
 The name of a resource is independent of the used operating system;
 in particular, the path separator is always a slash `/`.
+
+== Table Transformer
```
ikalinin1, line 51: I think it makes sense to put this reference somewhere here https://jbehave.org/reference/stable/tabular-parameters.html

ikalinin1 (approved):

uarlouski (approved):

docs/modules/developer-guides/pages/plugins.adoc
```diff
@@ -22,6 +22,99 @@ to the project root.
...
+Examples:
+{transformer=MY_TRANSFORMER}
+|header |
+|value 1|
+|value 2|
+----
+
+[NOTE]
+====
+It's recommended to stick to the following naming conventions:
```
uarlouski, line 88: can we also ask user to not create transformers with names defined in vvds?

PR #2712, opened 2022-04-29
[vividus] Add expression to quote regex patterns

ikalinin1:


docs/modules/commons/pages/expressions.adoc
```diff
@@ -323,6 +323,31 @@ Escapes reserved characters in https://www.w3schools.com/html/html_entities.asp[
 |`M&amp;Ms`
 |===
 
+=== `quotePattern`
```
valfirst, line 326: why is it needed? why users can't quote regex manually?

valfirst (changes requested):

docs/modules/commons/pages/expressions.adoc
ikalinin1: We are using backend data as the expected  value. In our case contains won't fit because we need stricter checks. I am not sure about all the cases, maybe a "starts with" check might help

valfirst, line 326:
```suggestion
=== `quoteRegExp`
```

valfirst (changes requested):

abudevich (approved):

valfirst (approved):

uarlouski (approved):

PR #2721, opened 2022-05-04
[agent-reportportal] Allow to disable attachments publishing

ikalinin1:


vividus-agent-reportportal/src/main/java/org/vividus/reportportal/config/condition/ReportPortalEnableCondition.java
```diff
@@ -18,16 +18,10 @@
...
-
-public class ReportPortalEnableCondition implements Condition
+public class ReportPortalEnableCondition extends AbstractBooleanCondition
 {
-    @Override
-    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata)
+    public ReportPortalEnableCondition()
     {
-        String enableProperty = ListenerProperty.ENABLE.getPropertyName();
-        return context.getEnvironment().getProperty(enableProperty, boolean.class);
```
uarlouski, line 31: as per javadoc "Return the property value associated with the given key, or null if the key cannot be resolved." I think it might lead to unexpected result in case of missing property

vividus-agent-reportportal/src/main/java/org/vividus/reportportal/config/condition/AbstractBooleanCondition.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.vividus.reportportal.config.condition;
+
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+public abstract class AbstractBooleanCondition implements Condition
```
uarlouski, line 23: mabe AbstractBooleanPropertyCondition? just AbstractBooleanCondition imho is toooo abstract

vividus-agent-reportportal/src/main/java/org/vividus/reportportal/config/condition/ReportPortalEnableCondition.java
ikalinin1: I think it is not the case, because we have the default value

docs/modules/integrations/pages/reportportal.adoc
```diff
@@ -47,19 +47,25 @@ a|`true`
...
 |The name of the launch, e.g. `SMOKE`
 
-|system.rp.description
+|*system.rp.description*
 |Any string
 |none
 |Launch description
 
-|system.rp.test-entity
+|*system.rp.test-entity*
```
valfirst, line 55: it's not mandatory since we have default value

docs/modules/integrations/pages/reportportal.adoc
```diff
@@ -47,19 +47,25 @@ a|`true`
...
 |`SCENARIO`/`STEP`
 |`SCENARIO`
 |Defines what will be reported as a test case `SCENARIO` or `STEP`
 
-|system.rp.attributes
+|*system.rp.attributes*
```
valfirst, line 60: is it really mandatory?

docs/modules/integrations/pages/reportportal.adoc
```diff
@@ -47,19 +47,25 @@ a|`true`
...
 |Semi-colon seprated values, e.g. `key:value; value;`
 |none
 |Attributes could be used for additional tagging of the launches
 
+|*system.rp.publish-attachments*
```
valfirst, line 65: it's not mandatory since we have default value

valfirst (changes requested):

valfirst (approved):

uarlouski (approved):

PR #2727, opened 2022-05-06
[visual-testing] Record assertion instead of exception when search co…

skyoker:
…ntext is null

### DIFF
--- a/vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
+++ b/vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
@@ -17,6 +17,7 @@
 package org.vividus.visual.steps;
 
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Supplier;
 
@@ -42,13 +43,14 @@ public abstract class AbstractVisualSteps
         this.softAssert = softAssert;
     }
 
-    protected <T extends VisualCheck> VisualCheckResult execute(Function<T, VisualCheckResult> checkResultProvider,
+    protected <T extends VisualCheck> Optional<VisualCheckResult> execute(
+            Function<T, VisualCheckResult> checkResultProvider,
             Supplier<T> visualCheckFactory, String templateName)
     {
         SearchContext searchContext = uiContext.getSearchContext();
         if (!softAssert.assertNotNull("Search context is set", searchContext))
         {
-            return null;
+            return Optional.empty();
         }
         T visualCheck = visualCheckFactory.get();
         visualCheck.setSearchContext(searchContext);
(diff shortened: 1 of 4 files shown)


vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
```diff
@@ -47,7 +46,10 @@ protected <T extends VisualCheck> VisualCheckResult execute(Function<T, VisualCh
             Supplier<T> visualCheckFactory, String templateName)
     {
         SearchContext searchContext = uiContext.getSearchContext();
-        Validate.validState(searchContext != null, "Search context is null, please check is browser session started");
+        if (!softAssert.assertNotNull("Search context is set", searchContext))
+        {
+            return null;
```
valfirst, line 51: it's better to return `Optional` instead

valfirst (commented):
Also please make sure the build is passed

vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
```diff
@@ -43,11 +43,15 @@ protected AbstractVisualSteps(IUiContext uiContext, IAttachmentPublisher attachm
         this.softAssert = softAssert;
     }
 
-    protected <T extends VisualCheck> VisualCheckResult execute(Function<T, VisualCheckResult> checkResultProvider,
+    protected <T extends VisualCheck> Optional<VisualCheckResult> execute(
```
valfirst, line 46: hmm, according to your changes, the result of this method is not used, so the better option would be to change option so that it returns `void`

ps: sorry for confusion, I've not checked the code before putting my comment about `Optional`

vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
```diff
@@ -43,11 +43,15 @@ protected AbstractVisualSteps(IUiContext uiContext, IAttachmentPublisher attachm
...
+            Function<T, VisualCheckResult> checkResultProvider,
             Supplier<T> visualCheckFactory, String templateName)
     {
         SearchContext searchContext = uiContext.getSearchContext();
-        Validate.validState(searchContext != null, "Search context is null, please check is browser session started");
```
ikalinin1, line 50:
```
return Optional.ofNullable(uiContext.getSearchContext(SearchContext.class)).map(sc -> {
    T visualCheck = visualCheckFactory.get();
        visualCheck.setSearchContext(searchContext);
       return visualCheck;
}).map(checkResultsProvider::apply).map(r -> {
           attachmentPublisher.publishAttachment(templateName, Map.of("result", result), "Visual comparison");
            verifyResult(result);
return result;
})
```

valfirst (approved):

uarlouski (approved):

vividus-extension-visual-testing/src/main/java/org/vividus/visual/steps/AbstractVisualSteps.java
skyoker: For me original code looks more readable

ikalinin1: Yes, but there is already a method, which does validation in ui context, why not reuse it?

PR #2874, opened 2022-06-15
[plugin-azure-storage-account] Extract blob storage client factory

ikalinin1:


uarlouski (approved):

vividus-plugin-azure-storage-account/src/main/java/org/vividus/azure/storage/blob/StorageAccountClientFactory.java
```diff
@@ -0,0 +1,74 @@
...
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+
+import org.apache.commons.lang3.Validate;
+import org.vividus.azure.storage.StorageAccountEndpoint;
+import org.vividus.util.property.PropertyMappedCollection;
+
+public class StorageAccountClientFactory
```
valfirst, line 32:
```suggestion
public class BlobServiceClientFactory
```

valfirst (changes requested):

valfirst (approved):

PR #2875, opened 2022-06-15
[engine] Support automatic conversion of bytes to string

ikalinin1:


vividus-engine/src/main/java/org/vividus/context/VariableTestContext.java
```diff
@@ -58,7 +59,13 @@ public void putVariable(Set<VariableScope> variableScopes, String variableKey, O
     public void putVariable(VariableScope variableScope, String variableKey, Object variableValue)
     {
         LOGGER.atInfo()
-                .addArgument(variableValue)
+                .addArgument(() -> {
+                    if (variableValue instanceof byte[])
+                    {
+                        return "Saving " + ((byte[]) variableValue).length / THOUSAND + "kb of binary data";
```
valfirst, line 65: it results in
`Saving a value 'Saving ... kb of binary data;' into ..`

valfirst, line 65: `kb` -> `kB`

valfirst (changes requested):

valfirst (approved):

PR #2914, opened 2022-06-29
[plugin-mobile-app] Fix invalid dpr calculation for android

uarlouski:


uarlouski:
> deviceScreenSize - is it always presented? Maybe handle corner cases?

it for sure exists for 4 years in both appium uiautomator2 and espresso drivers, if you have any thought about where corner cases may occur please share

ikalinin1:
> > deviceScreenSize - is it always presented? Maybe handle corner cases?
> 
> it for sure exists for 4 years in both appium uiautomator2 and espresso drivers, if you have any thought about where corner cases may occur please share

What about android tv? 
What about the real device?
Corner case is the case if the session doesn't have deviceScreenSize

valfirst (changes requested):

vividus-plugin-mobile-app/src/main/java/org/vividus/selenium/mobileapp/MobileAppWebDriverManager.java
```diff
@@ -124,12 +123,24 @@ public double getDpr()
...
+        {
+            String deviceScreenSize = (String) getCapabilities().getCapability("deviceScreenSize");
+            float screenHeight = Float.parseFloat(deviceScreenSize.split("x")[1]);
+            return calculateDpr(screenHeight);
+        }
+
+        return calculateDpr(getSize().getHeight());
+    }
+
+    private float calculateDpr(float screenHeight)
```
ikalinin1, line 137: Per discussion please update to use pixelRatio when available

ikalinin1 (approved):

valfirst (approved):

PR #2918, opened 2022-07-01
[plugin-mobile-app] Add step to send application to the background

### LINKED ISSUE #2841: [Mobile] Add possibility of sending application to the background
As a user, I want to be able to use a step that sends mobile application to the background. As a workaround emulation of key pressing is used for this purpose now. (e.g. Home);

Example of steps:
```
When I send application to the background for `$duration` seconds
```
**Additional context**
At the moment there is a vividus step for [activating application](https://docs.vividus.dev/vividus/latest/plugins/plugin-mobile-app.html#_activate_mobile_application).
[Appium command](https://appium.io/docs/en/commands/device/app/background-app/) for the functionality.

ikalinin1:
closes #2841

### DIFF
--- a/vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/steps/ApplicationSteps.java
+++ b/vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/steps/ApplicationSteps.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2021 the original author or authors.
+ * Copyright 2019-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import static io.appium.java_client.MobileCommand.prepareArguments;
 import static java.util.Map.entry;
 import static org.vividus.selenium.type.CapabilitiesValueTypeAdjuster.adjustType;
 
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -40,6 +41,7 @@ import org.vividus.selenium.manager.WebDriverManagerParameter;
 import org.vividus.util.property.PropertyParser;
 
 import io.appium.java_client.ExecutesMethod;
+import io.appium.java_client.InteractsWithApps;
 import io.appium.java_client.internal.CapabilityHelpers;
 
 public class ApplicationSteps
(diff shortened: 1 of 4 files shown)


vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/steps/ApplicationSteps.java
```diff
@@ -156,4 +158,15 @@ public void changeAppiumSettings(List<NamedEntry> settings)
 
         execute(executesMethod, entry("setSettings", prepareArguments("settings", prepareArguments(params, values))));
     }
+
+    /**
+     * Run application in background for given period of time.
+     * @param period  Total background run time according to
+     *                <a href="https://en.wikipedia.org/wiki/ISO_8601">ISO 8601</a> standard
+     */
+    @When("I run app in background for `$period` period")
```
uarlouski, line 167: In all steps that do something with the app we don't use shortcuts in their wording, but I also noticed that there are place with not just application, but mobile application, so I think we should standardise wordings for mobile app steps

vividus-tests/src/main/resources/story/system/mobile_app/MobileAppStepsTests.story
```diff
@@ -240,6 +240,19 @@ When I activate application with bundle identifier `${main-app}`
 When I wait until element located `accessibilityId(menuToggler)` appears
 
 
+Scenario: Verify step: 'When I run app in background for `$period` period'
+When I tap on element located `accessibilityId(menuToggler)`
+When I tap on element located `xpath(<menuScrollViewXpath>)`
+Then number of elements found by `accessibilityId(header)` is equal to `1`
+Then number of elements found by `accessibilityId(footer)` is equal to `0`
```
uarlouski, line 247: why not to check just header?

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/steps/ApplicationSteps.java
valfirst: `mobile application` -> 
```
When I send mobile application to background for `$period` period
```

valfirst (changes requested):

vividus-plugin-mobile-app/src/main/java/org/vividus/mobileapp/steps/ApplicationSteps.java
ikalinin1: @uarlouski  please create an issue to refactor existing steps

valfirst (approved):

uarlouski (approved):

PR #2926, opened 2022-07-05
[vividus-plugin-excel] Add ability to load excel table with different data types

### LINKED ISSUE #2908: Transformer FROM_EXCEL fails on importing .xlsx file with numbers
**Describe the bug**
Unable to use FROM_EXCEL transformer if the Excel file contains numbers; 
**To Reproduce**
Create bug.xlsx file and put it in the resource directory:
|theFirstSample|theSecondSample|number|
|----|---|---|
|pumpkin       |PUMPKIN        |7     |
|cucumber      |CUCUMBER       |8     |

Then run the following scenario:
```
Scenario: Test
When I initialize the SCENARIO variable `first` with value `<theFirstSample>`
Then `<theSecondSample>` is = `#{toUpperCase(${first})}`
Then `#{eval(size("<theSecondSample>"))}` is = `<number>`
Examples:
{transformer=FROM_EXCEL, path=bug.xlsx, sheet=Sheet, range=A1:C3}
```

**Expected behavior**
All table values are imported, test passes
**Actual behaviour**
Build fails with the error:
`ERROR org.vividus.runner.StoriesRunner - java.lang.IllegalStateException: Cannot get a STRING value from a NUMERIC cell`

**Execution environment (please complete the following information):**
 - Vividus version: 0.4.9
 - Java: 
 ```
java 16.0.2 2021-07-20
Java(TM) SE Runtime Environment (build 16.0.2+7-67)
Java HotSpot(TM) 64-Bit Server VM (build 16.0.2+7-67, mixed mode, sharing)
```

### PR DESCRIPTION
**Pull request by draker94**
Fixes #2908

### CODE REVIEW

File: vividus-tests/src/main/resources/story/integration/TableTransformers.story [hunk 1]
```diff
@@ -151,7 +151,18 @@ Meta:
     @issueId 647
 Then `<joined>` is equal to `line 1 line 2 line 3`
 Examples:
-{transformer=FROM_EXCEL, path=/data/excel.xlsx, sheet=Sheet1, addresses=A1, column=joined, \{lineBreakReplacement|VERBATIM\}= }
+{transformer=FROM_EXCEL, path=/data/excel.xlsx, sheet=Sheet1, addresses=A1, column=joined, \{lineBreakReplacement|VERBATIM\}= }\
+
+Scenario: Verify FROM_EXCEL transformer with different types of values
```
ikalinin1, line 156: Looks confusing to me, the user doesn't know anything about the types, and all the data is loaded as a string?

ikalinin1: This is a bugfix, right?

ikalinin1: Also, as an option you could combine:
When I initialize scopes variable `variableName` with values:examplesTable
Then `variable` is equal to table:table

ikalinin1 (approved):

valfirst (approved):

PR #2947, opened 2022-07-14
[plugin-web-app] Do not use Windows 10 as default platform

ikalinin1:


vividus-plugin-web-app/src/main/resources/properties/profile/web/desktop/profile.properties
```diff
@@ -1,7 +1,7 @@
 bdd.all-meta-filters=groovy: (!layout || layout.contains('desktop')) && !skip && (${bdd.meta-filters})
 
-selenium.grid.platform-name=Windows
-selenium.grid.platform-version=10
+selenium.grid.platform-name=
+selenium.grid.platform-version=
```
uarlouski, line 4: should we reflect this in docs? since users relied on this implicitly

ikalinin1: I don't think so, nevertheless it makes sense to highlight it in 0.5.0 release docs

valfirst, line 4: which platform is used in case of empty values by:
- SauceLabs
- BrowserStack
- LambdaTest
- custom Selenium Grid
?

valfirst, line 4: also it makes sense to add docs for these properties

valfirst (changes requested):

vividus-plugin-web-app/src/main/resources/properties/profile/web/desktop/profile.properties
ikalinin1: I believe this is a dynamic value and will change over time.  For now it seems:
LT: 10
BS: 8.1 (quite strange)
SL: 7 (veeeery strange)

I don't get the question about the custom grid

vividus-plugin-web-app/src/main/resources/properties/profile/web/desktop/edge/chromium/profile.properties
```diff
@@ -1,6 +1,5 @@
 selenium.browser=edge_chromium
 
-selenium.grid.capabilities.platformName=ANY
```
valfirst, line 3: isn't Windows required for legacy Edge as well?

.github/workflows/gradle.yml
```diff
@@ -153,6 +153,8 @@ jobs:
                                                   -Pvividus.configuration.profiles=browserstack/web,web/desktop/chrome \
                                                   -Pvividus.selenium.grid.username=${BROWSERSTACK_USER} \
                                                   -Pvividus.selenium.grid.password=${BROWSERSTACK_KEY} \
+                                                  -Pvividus.selenium.grid.platform-name=Windows \
+                                                  -Pvividus.selenium.grid.platform-version=10 \
```
valfirst, line 157: does it make sense to put these values to `browserstack/web` profile with comment saying BS uses by default very old windows, so we need to use something more relevant?

ikalinin1: I don't think we need to do this:
1. Default OS will change for sure in the future.
2.  The OS version is not really bound to the browser, for example you can run edge chromium on macOS
image

valfirst (approved):

uarlouski (approved):

PR #2966, opened 2022-07-20
[plugin-visual] Fix azure baseline storage condition

ikalinin1:


vividus-plugin-visual/src/main/java/org/vividus/visual/storage/AzureBlobStorageBaselineStorage.java
```diff
@@ -19,25 +19,33 @@
...
+@Conditional(BlobStorageCondition.class)
 public class AzureBlobStorageBaselineStorage implements BaselineStorage
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(AzureBlobStorageBaselineStorage.class);
 
     private final BlobServiceClientFactory blobServiceClientFactory;
 
+    @Value("${ui.visual.baseline-storage.azure-blob-storage.container:}")
     private String container;
+    @Value("${ui.visual.baseline-storage.azure-blob-storage.account-key:}")
```
uarlouski, line 48: since now this bean is conditional do we still need default values in placeholders?

ikalinin1: Yep, because you may still not use the storage provider even if the condition matching

valfirst (approved):

uarlouski (approved):

PR #3028, opened 2022-08-04
[vividus] Implement 'loadFile' and 'loadBinaryFile' predefined expres…

IBrun:
…sions


valfirst:
@IBrun why are these expressions required? why can't resources or steps be used? I do not support the idea to have a lot of interactions with file system as it will bring a lot of complexity for users in understanding the concept

PR #3035, opened 2022-08-08
[docs] Document steps сhecking for element disappearance

draker94:


valfirst (approved):

uarlouski (approved):

vividus-extension-selenium/src/main/java/org/vividus/ui/action/ExpectedSearchActionsConditions.java
```diff
@@ -32,7 +32,6 @@ public final class ExpectedSearchActionsConditions extends AbstractExpectedCondi
     @Override
     protected List<WebElement> findElements(SearchContext searchContext, Locator locator)
     {
-        locator.getSearchParameters().setWaitForElement(false);
```
ikalinin1, line 35: I don't think that's correct to remove it from here.
The usage of the waiter here involves the application of presenceAllElements condition.
We currently have this usage of the same invisibility condition
image

uarlouski: This is neither correct place to have this invocation since this toooo generic to manipulate locator parameters, its your concern that for each invocation of presenceOfAllElementsLocatedBy in steps we should mutate input locator?

ikalinin1: 1. If you feel lucky you could change search actions api to allow search without default waiting
2. In this implementation, every single condition receives presenceOfAllElementsLocatedBy by default. It is not transparent nor correct.

ikalinin1 (approved):

valfirst (approved):

uarlouski (approved):

PR #3078, opened 2022-08-18
[vividus-plugin-web-app] Improve search checkboxes by labels

vkepin:


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
```diff
@@ -77,7 +80,10 @@ private List<WebElement> searchCheckboxByLabels(SearchContext searchContext, Sea
             }
             else
             {
-                checkboxes = label.findElements(getXPathLocator(PRECEDING_SIBLING_CHECKBOX_LOCATOR));
+                checkboxes = Stream
+                        .of(PRECEDING_SIBLING_CHECKBOX_LOCATOR, FOLLOWING_SIBLING_CHECKBOX_LOCATOR,
+                                CHILD_CHECKBOX_LOCATOR)
```
valfirst, line 85: do we really need to find checkboxes by all locators? won't the first one be enough?

valfirst (changes requested):
Please do not forget to add integration tests

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
```diff
@@ -77,7 +80,10 @@ private List<WebElement> searchCheckboxByLabels(SearchContext searchContext, Sea
             }
             else
             {
-                checkboxes = label.findElements(getXPathLocator(PRECEDING_SIBLING_CHECKBOX_LOCATOR));
+                checkboxes = Stream
+                        .of(PRECEDING_SIBLING_CHECKBOX_LOCATOR, FOLLOWING_SIBLING_CHECKBOX_LOCATOR,
```
ikalinin1, line 84: .map(this::getXpathLocator).map(label::findElements)

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
```diff
@@ -77,7 +80,10 @@ private List<WebElement> searchCheckboxByLabels(SearchContext searchContext, Sea
             }
             else
             {
-                checkboxes = label.findElements(getXPathLocator(PRECEDING_SIBLING_CHECKBOX_LOCATOR));
+                checkboxes = Stream
```
ikalinin1, line 83: What about integration tests update? :)

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
vkepin: no, because with CHECKBOX_LABEL_FORMAT we find for example child element, but then lost it in this method
image

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
vkepin: integration test should work fine with this change ;)

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
vkepin: here should be flatMap, as I merge 3 lists for each locator in 1. I can prepare locators in map if it's really necessary (not sure)

ikalinin1: I think it's a rules of thumb: use method references where applicable, perform operations atomically on stream elements where applicable

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
vkepin: on the other hand, I can change logic and use 'for' cycle and exit once first element collection will be find. What do you think?

valfirst: > exit once first element collection will be find

yes, that I was asking about

ikalinin1: java.util.stream.Stream#findFirst ??

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/CheckboxNameSearch.java
valfirst: that's true, but we should add integration tests to cover new logic

vkepin:
Ok, I'll just change locator in my tests for now. I'll return once I have enough time to work on this PR

PR #3099, opened 2022-08-25
[exporter] Consider assertion failures for test status calculation

uarlouski:
Assertion failures go into system steps
Brokens go into user steps

### DIFF
--- a/vividus-engine/src/main/java/org/vividus/model/jbehave/AbstractStepsContainer.java
+++ b/vividus-engine/src/main/java/org/vividus/model/jbehave/AbstractStepsContainer.java
@@ -21,9 +21,21 @@ import java.util.stream.Stream;
 
 public abstract class AbstractStepsContainer
 {
+    private List<Step> beforeSystemScenarioSteps;
     private List<Step> beforeUserScenarioSteps;
     private List<Step> steps;
     private List<Step> afterUserScenarioSteps;
+    private List<Step> afterSystemScenarioSteps;
+
+    public List<Step> getBeforeSystemScenarioSteps()
+    {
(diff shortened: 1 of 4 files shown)


vividus-engine/src/main/java/org/vividus/model/jbehave/AbstractStepsContainer.java
```diff
@@ -55,8 +56,19 @@ public void setAfterUserScenarioSteps(List<Step> afterScenarioSteps)
...
+    public void setAfterSystemScenarioSteps(List<Step> afterSystemScenarioSteps)
+    {
+        this.afterSystemScenarioSteps = afterSystemScenarioSteps;
+    }
+
     public Stream<Step> createStreamOfAllSteps()
     {
-        return Stream.of(getBeforeUserScenarioSteps(), getSteps(), getAfterUserScenarioSteps()).flatMap(List::stream);
+        return Stream.of(getBeforeUserScenarioSteps(), getSteps(), getAfterUserScenarioSteps(),
+                getAfterSystemScenarioSteps()).flatMap(List::stream);
```
valfirst, line 72: why not to add `beforeSystemScenarioSteps`?

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/model/jbehave/AbstractStepsContainer.java
uarlouski: there is nothing that can be failed

ikalinin1 (approved):

valfirst (approved):

PR #3151, opened 2022-09-13
[plugin-web-app] Remove deprecated code steps

ikalinin1:

PR #3153, opened 2022-09-13
[plugin-excel] Remove deprecated Excel validating steps

ikalinin1:


ikalinin1:
#3151 [plugin-excel] Remove deprecated Excel validating steps

|Deprecated step|Replacement|
|--|--|
|Then response contains excel sheet with index \`$index\` and records:$records|Then \`${response-as-bytes}\` contains excel sheet with index \`\<index\>\` and records:\<records\>|
|Then response contains excel sheet with name \`$name\` and records:$records|Then \`${response-as-bytes}\` contains excel sheet with name \`\<name\>\` and records:\<records\>|

valfirst (approved):

PR #3156, opened 2022-09-14
Standardize plugins configuration

uarlouski:


docs/modules/developer-guides/pages/plugins.adoc
```diff
@@ -3,8 +3,9 @@
 Plugins are extensible and flexible mechanism for development. New plugin can be developed by
 3rd party team and integrated.
 
-The plugin configuration should be placed at the `/src/main/resources/spring.xml` path relatively
-to the project root.
+The configuration for custom plugins that won't be integrated into VIVIDUS due to their specificity should be placed at the `/src/main/resources/spring.xml` path
```
ikalinin1, line 6: please, renovate example of the xml to highlight the changes.

ikalinin1, line 6: The difference between locations is unclear, maybe add some specific examples?

valfirst: I would say we need to introduce the following naming: 
 - extensions - basement for plugins
 - plugins - contain generic features that can be used by multiple projects, it could be either VIVIDUS or external ones
 - custom modules - contain project-specific logic (steps, transformers, expressions, etc.), may depend on extensions and/or plugins, may not have dependencies at all, also they may override default behaviour of plugins and extensions if needed

valfirst: this classification can be put to the extension points page and should be duplicated for each component

valfirst (approved):

PR #3172, opened 2022-09-19
Introduce YAML plugin

valfirst:


docs/modules/plugins/pages/plugin-yaml.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+* `$yaml` - The YAML used to find YAML element value.
+* `$yamlPath` - The {yaml-path} used to find YAML element value.
+* `$scopes` - xref:commons:variables.adoc#_scopes[The comma-separated set of the variables scopes].
+* `$variableName` - The name of the variable to save the found YAML element value.
+
+.Validate the title of the second book
+[source,gherkin]
+----
+When I save YAML element value from `${response}` by YAML path `store.book[1].title` to scenario variable `title-of-second-book`
```
uarlouski, line 40: maybe It makes sense to add an example of har you work with in this example

valfirst: - I don't have any working URL that hosts public YAML
- adding YAMLs to examples will significantly increase the size of the docs

ikalinin1 (approved):

uarlouski (approved):

PR #3201, opened 2022-09-26
[extension-selenium] Search for element at first in step validating element existence

### LINKED ISSUE #2750: "Then element located ... exists for ... duration" fails when element does not appear yet
**Describe the bug**
The following step:
```
Then element located `locator` exists for `duration` duration
```
fails immediately in case element did not have chance to appear on the screen.

**To Reproduce**
Apply provided step to an element which is not on the screen at the beginning of scenario (e.g. toast message)

**Expected behavior**
Step waits for the element appearance and then checks if this element exists for expected duration.

**Execution environment (please complete the following information):**
 - Vividus version: 0.4.8-SNAPSHOT

valfirst:
Closes #2750
Supersedes #3035

### DIFF
--- a/vividus-extension-selenium/src/main/java/org/vividus/steps/ui/GenericWaitSteps.java
+++ b/vividus-extension-selenium/src/main/java/org/vividus/steps/ui/GenericWaitSteps.java
@@ -23,8 +23,10 @@ import org.apache.commons.lang3.time.DurationFormatUtils;
 import org.jbehave.core.annotations.Then;
 import org.jbehave.core.annotations.When;
 import org.vividus.softassert.ISoftAssert;
+import org.vividus.steps.ui.validation.IBaseValidations;
 import org.vividus.ui.action.IExpectedConditions;
 import org.vividus.ui.action.IExpectedSearchContextCondition;
+import org.vividus.ui.action.ISearchActions;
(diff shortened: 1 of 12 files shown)


vividus-extension-selenium/src/main/java/org/vividus/steps/ui/GenericWaitSteps.java
```diff
@@ -69,23 +76,40 @@ public void waitForElementDisappearance(Locator locator)
...
+     * @param locator  The locator to find an element.
+     * @param duration The period of time the element should have existed in
+     *                 <a href="https://en.wikipedia.org/wiki/ISO_8601">ISO 8601</a>
+     *                 format.
      */
     @Then("element located `$locator` exists for `$duration` duration")
     public void doesElementByLocatorExistsForDuration(Locator locator, Duration duration)
     {
-        String prettyDuration = formatDuration(duration);
+        String prettyPrintedDuration = formatDuration(duration);
```
uarlouski, line 96: inline?

valfirst: we want to fail fast in case invalid duration (see the case from the unit tests)

ikalinin1 (approved):

PR #3211, opened 2022-09-28
Refactor steps to work with JSON

valfirst:


docs/modules/plugins/pages/plugin-json.adoc
```diff
@@ -10,24 +10,52 @@ The plugin provides a set of actions for transformation and validation of JSON d
...
    (`true` or `false`) and `null`.
 
-NOTE: JSON context is pointed to the latest HTTP response by default.
-
 === Save JSON element value from context
 
-Saves a value of JSON element found in JSON context into the variable with
+[WARNING]
+====
+The step is deprecated and will be removed in VIVIDUS 0.6.0. The replacement is
```
uarlouski, line 48: why do we need deprecated step docs?

valfirst: yes,
- what if someone uses this step, how to understand what it does
- what if there is a link to this doc, we should keep it for a while

uarlouski: imho
- if the steps is already used in scenario then it's faster to check it in javadocs
- if end user is not aware of this step and has intent to save a value from JSON he will probably open the docs and find new step by keywords

uarlouski (approved):

ikalinin1 (approved):

PR #3231, opened 2022-10-07
Fix `OutOfMemoryError` on local builds using Gradle Daemon

valfirst:


gradle.properties
```diff
@@ -3,3 +3,4 @@ buildSystemRootDir=vividus-build-system
 
 org.gradle.parallel=true
 org.gradle.caching=true
+org.gradle.jvmargs=-Xmx2g
```
ikalinin1, line 6: looks like a workaround :)

ikalinin1 (approved):

uarlouski (approved):

gradle.properties
valfirst: that's true, I assume there is some issue with SpotBugs, but it's too expensive to debug it

PR #3349, opened 2022-11-14
[vividus] Fix handling of an attempt to run empty suite

valfirst:


vividus/src/main/java/org/vividus/runner/StoriesRunner.java
```diff
@@ -51,6 +51,10 @@ public StoriesRunner()
     public void run()
     {
         Map<String, List<String>> paths = getPaths();
+        if (paths.isEmpty())
+        {
+            throw new InvalidConfigurationException("No batches with tests to execute are configured");
```
ikalinin1, line 56: EmptySuites(Batches)Exception?

Can we consider the selenium way  https://www.selenium.dev/exceptions/

uarlouski (approved):

vividus/src/main/java/org/vividus/runner/StoriesRunner.java
uarlouski: just validate.isTrue?)

valfirst: > EmptySuites(Batches)Exception

I plan the generic exception will be reused.

> Can we consider the selenium way https://www.selenium.dev/exceptions/

No references to docs (yet) - too many potential issues, so again no reason to make this exception very specific.

> just validate.isTrue?)

`IllegalArgumentException` is too generic and used too often

uarlouski: I bet end users won't check the exception name

valfirst: Some specific exception is easier to find it in the docs

ikalinin1 (approved):

PR #3357, opened 2022-11-16
[vividus] Allow handling of empty tables to CARTESIAN_PRODUCT transformer according to mathematical logic

draker94:
[Cartesian Product of Empty Set](https://www.storyofmathematics.com/empty-set/#:~:text=The%20Cartesian%20Product%20of%20any,empty%20set%20contains%20no%20elements.)

### DIFF
--- a/docs/modules/commons/pages/table-transformers.adoc
+++ b/docs/modules/commons/pages/table-transformers.adoc
@@ -449,7 +449,7 @@ Examples:
 ====
 * headers of input tables must be different
 * number of input tables must be more than 1
-* empty tables are not allowed
+* if any of tables is empty - the resulting table will also be empty
 ====
 
 .Usage of CARTESIAN_PRODUCT transformer
--- a/vividus-tests/src/main/resources/story/integration/TableTransformers.story
+++ b/vividus-tests/src/main/resources/story/integration/TableTransformers.story
@@ -263,6 +263,11 @@ Then `${cartesianProductTable}` is equal to table:
 |3 |AWS    |Neptune |row21|row22|row23|
 |3 |AWS    |Neptune |row31|row32|row33|
 
+Scenario: Verify CARTESIAN_PRODUCT transformer with empty table (should not be executed)
+Then `should` is equal to `be skipped`
+Examples:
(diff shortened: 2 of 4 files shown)


docs/modules/commons/pages/table-transformers.adoc
```diff
@@ -449,7 +449,7 @@ Examples:
 ====
 * headers of input tables must be different
 * number of input tables must be more than 1
-* empty tables are not allowed
+* if one table is empty - the transformed table will also be empty
```
valfirst, line 452:
```suggestion
* if any of tables is empty - the resulting table will also be empty
```

vividus-tests/src/main/resources/story/integration/TableTransformers.story
```diff
@@ -263,6 +263,11 @@ Then `${cartesianProductTable}` is equal to table:
 |3 |AWS    |Neptune |row21|row22|row23|
 |3 |AWS    |Neptune |row31|row32|row33|
 
+Scenario: Verify CARTESIAN_PRODUCT transformer with empty table (should be skipped)
```
valfirst, line 266:
```suggestion
Scenario: Verify CARTESIAN_PRODUCT transformer with empty table (should not be executed)
```

to avoid collision with a special state `skipped` for steps/scenarios

vividus/src/main/java/org/vividus/transformer/CartesianProductTableTransformer.java
```diff
@@ -60,6 +61,11 @@ public String transform(String tableAsString, TableParsers tableParsers, TablePr
         isTrue(repeatingKeys.isEmpty(), "Tables must contain different keys, but found the same keys: %s",
                 repeatingKeys);
 
+        if (hasEmptyTable(tables))
+        {
+            return getEmptyTableWithHeaders(tables, properties);
+        }
```
valfirst, line 67: what will happen with empty table(s) without this `if` (if they go through the logic below)?

valfirst (changes requested):

valfirst (approved):

ikalinin1 (approved):

uarlouski (approved):

PR #3380, opened 2022-11-22
[plugin-datetime] Refactor expression processors

EDbarvinsky:
Relates #3256

### DIFF
--- a/vividus-plugin-datetime/src/main/java/org/vividus/expression/DateExpression.java
+++ b/vividus-plugin-datetime/src/main/java/org/vividus/expression/DateExpression.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2021 the original author or authors.
+ * Copyright 2019-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,11 +19,13 @@ package org.vividus.expression;
 import java.time.Duration;
 import java.time.Period;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
 import java.util.regex.Matcher;
 
 import org.apache.commons.lang3.StringUtils;
 
-public final class DateExpression
+public final class DateExpression implements NormalizingArguments
 {
     private static final String DURATION_DESIGNATOR = "P";
 
@@ -39,6 +41,26 @@ public final class DateExpression
         durationString = matcher.group(durationGroup);
         customFormatString = matcher.group(formatGroup);
         minusSign = matcher.group(minusSignGroup);
+    }
+
+    public String format(ZonedDateTime zonedDateTime, Locale locale)
+    {
+        DateTimeFormatter format = DateTimeFormatter.ISO_LOCAL_DATE;
+        ZonedDateTime result = zonedDateTime;
+        if (hasPeriod())
+        {
+            result = processPeriod(result);
+        }
+        if (hasDuration())
+        {
+            result = processDuration(result);
+            format = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
+        }
+        if (hasCustomFormat())
+        {
+            format = DateTimeFormatter.ofPattern(getCustomFormatString(), locale);
+        }
+        return result.format(format);
     }
 
     public boolean hasPeriod()
@@ -58,7 +80,7 @@ public final class DateExpression
 
     public String getCustomFormatString()
     {
-        return customFormatString;
+        return normalize(customFormatString);
     }
 
     public ZonedDateTime processPeriod(ZonedDateTime zonedDateTime)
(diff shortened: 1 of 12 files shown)


vividus-plugin-datetime/src/main/java/org/vividus/expression/TemporalFormatter.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.vividus.expression;
+
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Locale;
+
+interface TemporalFormatter
+{
+    default String format(DateExpression dateExpression, ZonedDateTime zonedDateTime, Locale locale)
```
EDbarvinsky, line 25: looks like could be moved to the `DateExpression`

ikalinin1 (approved):

vividus-plugin-datetime/src/main/java/org/vividus/expression/ShiftDateExpressionProcessor.java
```diff
@@ -57,19 +55,7 @@ protected String evaluateExpression(Matcher expressionMatcher)
...
-        {
-            zonedDateTime = dateExpression.processDuration(zonedDateTime);
-        }
-        return format.format(zonedDateTime);
-    }
-
-    private String normalize(String argument)
-    {
-        return StringUtils.replace(argument.trim(), "\\,", ",");
+        String formattedDateTime = dateExpression.format(zonedDateTime, Locale.ENGLISH);
```
valfirst, line 58: move Locale to constant

vividus-plugin-datetime/src/main/java/org/vividus/expression/ShiftDateExpressionProcessor.java
```diff
@@ -57,19 +55,7 @@ protected String evaluateExpression(Matcher expressionMatcher)
...
+        return normalize(formattedDateTime);
```
valfirst, line 59: why is result normalized?

valfirst (changes requested):

vividus-plugin-datetime/src/main/java/org/vividus/expression/ShiftDateExpressionProcessor.java
EDbarvinsky: my mistake. Previously was used  `DateTimeFormatter format = DateTimeFormatter.ofPattern(normalize(expressionMatcher.group(FORMAT_GROUP)),` which is normalize, but as logic for formatiing moved to the `DateExpression` it is also should be normalized inside it. fixed

vividus-plugin-datetime/src/main/java/org/vividus/expression/DateExpression.java
```diff
@@ -41,6 +43,26 @@ public DateExpression(Matcher matcher, int minusSignGroup, int periodGroup, int
...
+        DateTimeFormatter format = DateTimeFormatter.ISO_LOCAL_DATE;
+        ZonedDateTime result = zonedDateTime;
+        if (hasPeriod())
+        {
+            result = processPeriod(result);
+        }
+        if (hasDuration())
+        {
+            result = processDuration(result);
+            format = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss");
```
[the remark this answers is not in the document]
EDbarvinsky: replaced with [ISO_LOCAL_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE_TIME)

vividus-plugin-datetime/src/main/java/org/vividus/expression/DateExpression.java
```diff
@@ -58,7 +80,7 @@ public boolean hasCustomFormat()
 
     public String getCustomFormatString()
     {
-        return customFormatString;
+        return normalize(customFormatString);
```
EDbarvinsky, line 83: [getPeriod](https://github.com/vividus-framework/vividus/pull/3380/commits/7aa18209d7f81c7280b5cd44a443d89ff75e7ee4#diff-d1956a62e8bd9e3a84df60796619e587c6cac8724814db2df5344e602ce64441R110) loogicaly also should be normalized, for re-usability purpose, but at the same time according to the current implementation that does not make sense as input parameter for Period and Duration by it is meaning should not had `\\,` so that does not make sense to do, agree ?

EDbarvinsky:
> Would you mind updating docs to demo escaping features for different expressions?

updated where was required.

valfirst (changes requested):

vividus/src/test/java/org/vividus/expression/ReplaceByRegExpProcessorTests.java
```diff
@@ -41,6 +41,7 @@ static Stream<Arguments> expressionInput()
             of("formatDate()",                                                                      Optional.empty()),
             of("replaceFirstByRegExp(\"\"\".*three, (\\d)\"\"\", $1.0, \"\"\"two, three, 4\"\"\")", Optional.of("4.0")),
             of("replaceAllByRegExp(\\s, _, string with spaces)",                                    Optional.of("string_with_spaces")),
+            of("replaceAllByRegExp(,, \\,_\\,, string,with,comma)",                                 Optional.of("string,_,with,_,comma")),
```
valfirst, line 44: it seems the changes were not split correctly

valfirst (approved):

uarlouski (approved):

PR #3388, opened 2022-11-24
[engine] Add logging table transformer monitor

draker94:


vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void afterTransformerApplying(String transformerName, TableProperties properties, String outputTable)
+    {
+        logTransformerInfo(transformerName, properties, outputTable, "Output (transformed) table");
+    }
+
+    private void logTransformerInfo(String transformerName, TableProperties tableProperties, String table,
+            String tableType)
+    {
+        StringBuilder formatBuilder = new StringBuilder("Applying table transformer: {}").append(LINE_SEPARATOR)
```
ikalinin1, line 59: Why not choose the format instead of the 'builder and format'

vividus-engine/src/test/java/org/vividus/log/LoggingTableTransformerMonitorTests.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.jupiter.params.provider.CsvSource;
+
+class LoggingTableTransformerMonitorTests
+{
+    private static final String LINE_SEPARATOR = System.lineSeparator();
+    private static final String TRANSFORMER_NAME = "TEST_TRANSFORMER";
+    private static final String APPLYING_LINE = "Applying table transformer: {}";
+    private static final String PROPERTIES_LINE = "Properties: {}";
+    private static final String PARAMETER_LINE = "{}";
+    private static final String BEFORE_APPLYING_FORMAT = APPLYING_LINE + LINE_SEPARATOR
```
ikalinin1, line 44: From my point of view having plain strings would be much simpler to read :)

draker94: Yes, and originally it was, but the checkstyle said "no".
Perhaps it makes sense then to ignore the rule

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,94 @@
...
+                .append("Properties: {}").append(LINE_SEPARATOR)
```
valfirst, line 60: use native ExamplesTable notation: https://jbehave.org/reference/preview/tabular-parameters.html

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,94 @@
...
+        String tableForLogging = getExampleTableForLogging(table, tableProperties);
+
+        LOGGER.info(format, transformerName, propertiesForLogging, tableForLogging);
+    }
+
+    private String getExampleTableForLogging(String exampleTableOrigin, TableProperties tableProperties)
+    {
+        if (StringUtils.isBlank(exampleTableOrigin))
+        {
+            return "EMPTY";
```
valfirst, line 73:
```suggestion
            return ">>>  EMPTY  <<<";
```
to make it more unique

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        String tableForLogging = FORMATTING.transform(exampleTableOrigin, tableParsers, tableProperties).trim();
```
valfirst, line 76: I don't think we should format all rows in case if table is truncated

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        return tableForLogging;
+    }
+
+    private String truncateTable(String tableForLogging, String rowSeparator, long numOfRows)
+    {
+        String truncatedTable = tableForLogging.lines().limit(MAX_ROWS_LOGGING)
+                .collect(Collectors.joining(rowSeparator));
+        long rowsTruncated = numOfRows - MAX_ROWS_LOGGING;
+        return truncatedTable + String.format("%s...%d row(s) truncated", rowSeparator, rowsTruncated);
```
valfirst, line 92: no reason to use String concatenation and String.format in the single statement

vividus-engine/src/test/java/org/vividus/log/LoggingTableTransformerMonitorTests.java
```diff
@@ -0,0 +1,115 @@
...
+    private final TableParsers tableParsers = new TableParsers(parameterConverters);
+    private final LoggingTableTransformerMonitor monitor = new LoggingTableTransformerMonitor(tableParsers);
+    private final TableProperties tableProperties = new TableProperties("{transformer=TEST_TRANSFORMER}",
+            keywords, parameterConverters);
+
+    @BeforeEach
+    void beforeEach()
+    {
+        logger.clear();
+    }
```
valfirst, line 68: use `TestLoggerFactoryExtension` instead

vividus-engine/src/test/java/org/vividus/log/LoggingTableTransformerMonitorTests.java
valfirst: yep, just suppress it for this class

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,104 @@
...
+        logTransformerInfo(transformerName, properties, outputTable, loggerFormat);
+    }
+
+    private void logTransformerInfo(String transformerName, TableProperties tableProperties, String table,
+            String loggerFormat)
+    {
+        String propertiesForLogging = tableProperties.getPropertiesAsString().trim();
+        String tableForLogging = getExampleTableForLogging(table, tableProperties);
+
+        LOGGER.info(loggerFormat, transformerName, propertiesForLogging, tableForLogging);
```
uarlouski, line 65: there is lots of stuff going on with strings, but I think we should take into account a low chance that the info level might be disabled, so in this case, all these computations should not be performed

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,104 @@
...
+        long numOfRows = exampleTableOrigin.lines().count();
+        if (numOfRows > MAX_ROWS_LOGGING)
+        {
+            tableForLogging = truncateTable(exampleTableOrigin, tableProperties, numOfRows);
+        }
+        else
+        {
+            tableForLogging = formatTable(exampleTableOrigin, tableProperties);
+        }
+        return tableForLogging;
```
uarlouski, line 86: can we use just ternary here?

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,104 @@
...
+    {
+        String loggerFormat = String
+                .format("Applying table transformer: {}%nTabular parameters: {}%nOutput (transformed) table:%n{}");
+        logTransformerInfo(transformerName, properties, outputTable, loggerFormat);
+    }
+
+    private void logTransformerInfo(String transformerName, TableProperties tableProperties, String table,
+            String loggerFormat)
+    {
+        String propertiesForLogging = tableProperties.getPropertiesAsString().trim();
```
valfirst, line 62: I would expect the properties logged in the following format:
```
{headerSeparator=..., }
|k1|k2|
...
```
and they should be logged only if there values which differ from default ones

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
draker94: Did you mean add braces?
"getPropertiesAsString" returns by default only user properties
But I don't think that we must skip property section at all, because transformer is also a property (which can have its own properties)

draker94: @valfirst 
Since we don't have a strict separation between the properties of the table and the properties of the transformer I would not expect that we may not have properties
tableProps

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,89 @@
...
+import org.jbehave.core.model.TableTransformers.Formatting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingTableTransformerMonitor implements TableTransformerMonitor
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(LoggingTableTransformerMonitor.class);
+
+    private static final Formatting FORMATTING_TRANSFORMER = new Formatting();
+    private static final int MAX_ROWS_LOGGING = 50;
```
ikalinin1, line 35:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(LoggingTableTransformerMonitor.class);

    private static final Formatting FORMATTING_TRANSFORMER = new Formatting();
    
    private static final int MAX_ROWS_LOGGING = 50;
```

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        if (StringUtils.isBlank(originExampleTable))
+        {
+            return ">>>  EMPTY  <<<";
+        }
+
+        AtomicInteger totalNumberOrRows = new AtomicInteger();
+        String rowSeparator = tableProperties.getRowSeparator();
+        String resultingTable = originExampleTable.lines()
+                .filter(ignored -> totalNumberOrRows.incrementAndGet() <= MAX_ROWS_LOGGING)
```
ikalinin1, line 76:
```suggestion
                .limit(MAX_ROWS_LOGGING)
```

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
valfirst: the matter of taste

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
ikalinin1: image

ikalinin1: @valfirst  could you please elaborate?

ikalinin1: Oh I missed that :) Yet from my point of view if you need to count the number of joins it should be done by the collector. 
Yet the current implementation performance depends on the number of lines. What if we have ve_eee_eee_eee_ry long table :)

ikalinin1 (approved):

uarlouski (approved):

vividus-engine/src/main/java/org/vividus/log/LoggingTableTransformerMonitor.java
valfirst: the same was before, the difference is now table is parsed once

PR #3391, opened 2022-11-25
[vividus] Allow to escape commas in common expressions

EDbarvinsky:
Relates #3256


vividus/src/test/java/org/vividus/expression/HashExpressionProcessorTests.java
```diff
@@ -40,6 +40,7 @@
             // CHECKSTYLE:OFF
             "'calculateHash(SHA-1, a_tylmarande@gmail.com)', 025623becec96e8be6b88a7db6e0dc97fa0033ba",
             "'calculateHash(Sha-256, a_tylmarande@gmail.com)', b4e73a7c815bb0eff534ba8ef5f7dbfe9d4f51f449cd9b2ba87fc57ee9de5fc7",
+            "'calculateHash(SHA256, reg,exp,works)', 1ce5cd2ea218c8d28748216c4a7c9bb51fc4d8c4fafd8d91d4edb8b1f9bd06a4",
```
EDbarvinsky, line 43: looks risky to add normalization for `\\,` as there is no guarantee that such input was not used as valid argument before.
Will we leave it as is or also will add normalization support ?

ikalinin1 (approved):

ikalinin1 (approved):

uarlouski (approved):

PR #3414, opened 2022-12-01
[vividus] Allow to set fail fast at batch level

ikalinin1:


vividus-tests/src/main/resources/story/system/mobile_app/MobitruHealthCheckMobileApp.story
```diff
@@ -16,3 +16,6 @@ Examples:
 Scenario: Validate application start
 When I wait until element located by `accessibilityId(test-Username)` appears
 When I take screenshot
+When I type `text` in field located `locator`
+When I press key key
+When I type `text` in field located `locator` and keep keyboard opened
```
valfirst, line 21: is it expected change?

vividus-engine/src/main/java/org/vividus/batch/BatchConfiguration.java
```diff
@@ -130,4 +152,24 @@ private List<String> convertToList(String list)
                      .map(String::strip)
                      .collect(Collectors.toList());
     }
+
+    private static final class Story
```
valfirst, line 156:
```suggestion
    private static final class StoryExecutionConfiguration
```

vividus-engine/src/main/java/org/vividus/batch/BatchConfiguration.java
```diff
@@ -130,4 +152,24 @@ private List<String> convertToList(String list)
...
+    {
+        private boolean failFast;
+
+        private void setFailFast(boolean failFast)
+        {
+            this.failFast = failFast;
+        }
+    }
+
+    private static final class Scenario
```
valfirst, line 166:
```suggestion
    private static final class ScenarioExecutionConfiguration
```

valfirst (changes requested):

valfirst (approved):

vividus-engine/src/main/java/org/vividus/batch/BatchConfiguration.java
```diff
@@ -130,4 +152,24 @@ private List<String> convertToList(String list)
...
+
+    private static final class ScenarioExecutionConfiguration
+    {
+        private boolean failFast;
+
+        private void setFailFast(boolean failFast)
+        {
+            this.failFast = failFast;
+        }
+    }
```
uarlouski, line 174: minor question, do we need both StoryExecutionConfiguration and ScenarioExecutionConfiguration? it look like we could use one class with generic name

uarlouski (approved):

vividus-engine/src/main/java/org/vividus/batch/BatchConfiguration.java
ikalinin1: Based on the issue https://github.com/vividus-framework/vividus/issues/3416 this value will be inserted into the story config

PR #3429, opened 2022-12-05
[vividus] Do not use bdd prefix for variables definitions

ikalinin1:


docs/modules/commons/pages/variables.adoc
```diff
@@ -65,13 +65,13 @@ NOTE: The scopes names are case-insensitive.
 |Variables are available in the next batches. `next_batches`-scoped variable is not available in batch it's declared within.
 
 |`batch`
-a|Such variables are declared only with `bdd.variables.batch-<batch-number>.` prefix and available throughout the execution of the batch in the suite. An attempt to create the batch-scoped variables during test execution leads to an error.
+a|Such variables are declared only with `variables.batch-<batch-number>.` prefix and available throughout the execution of the batch in the suite. An attempt to create the batch-scoped variables during test execution leads to an error.
```
valfirst, line 68: logically it should be `batch-<batch-number>.variables.`, but I would say it's under discussion

ikalinin1: Really good point, I don't have objections!

valfirst: also in this case: `variables.global.` -> `variables.` (full sync between suite and batch levels)

vividus/src/main/resources/properties/deprecated/depreacted.properties
```diff
@@ -3,3 +3,5 @@ bdd.story.fail-fast=story.fail-fast
 bdd.scenario.fail-fast=scenario.fail-fast
 bdd.story-loader.batch-(\\d+)(.+)=batch-$1$2
 bdd.batch-(\\d+)(.+)=batch-$1$2
+bdd.variables.global(.+)=variables$1
```
valfirst, line 6:
```suggestion
bdd.variables.global.(.+)=variables.$1
```

valfirst (changes requested):

vividus/src/main/resources/properties/deprecated/depreacted.properties
uarlouski: btw should not we escape dots?

valfirst: yes, good cacth

valfirst (commented):
`../depreacted.properties` -> `../deprecated.properties`

PR #3433, opened 2022-12-07
Unify parsing of expression arguments

### LINKED ISSUE #3256: Make sure all VIVIDUS expressions allow to escape commas
All VIVIDUS expressions having 2 and more arguments should support escaping of commas:
```
#{expression(arg\,arg 1, arg\,2,  arg \, 3)}
```
- argument 1: `arg,arg 1` 
- argument 2: `arg,2`
- argument 3:  `arg , 3`

Pay attention: the leading whitespace for the 2-nd and subsequent arguments are stripped.

The documentation should be updated accordingly.

### LINKED ISSUE #3257: Make sure all VIVIDUS expressions can accept arguments as is
Prerequisite: #3256

Since the arguments of the expression are comma-separated, meaningful commas in arguments must be escaped with `\` (see #3256). But it might be not convenient when arguments are get dynamically (e.g. a part of HTTP response), in this case escaping could be a bit complicated. In order to handle such cases VIVIDUS expressions should have the ability to accept arguments as is: the argument must be wrapped into triple quotes: `"""arg value"""`
```
#{expression( """arg,arg 1 """, """arg,2""", """ arg , 3""")}
```
- argument 1: `arg,arg 1 ` 
- argument 2: `arg,2`
- argument 3:  ` arg , 3`

Pay attention: the leading whitespaces outside triple quotes are stripped, but leading and trailing whitespaces inside triple quotes are preserved.

The documentation should be updated accordingly.

EDbarvinsky:
Closes #3256
Closes #3257

- [x] Implementation
- [x] Unit and integration tests
- [x] Docs
- [x] Update #3257 references table


valfirst (commented):
Please check codecov comments spotting code branches which are not covered by unit tests

vividus-plugin-datetime/src/main/java/org/vividus/expression/DateExpression.java
```diff
@@ -80,7 +80,7 @@ public boolean hasCustomFormat()
 
     public String getCustomFormatString()
     {
-        return normalize(customFormatString);
+        return NormalizingArgumentsUtils.normalize(customFormatString);
```
valfirst, line 83: for all such changes it's better to use static import

vividus-engine/src/test/java/org/vividus/expression/NormalizingArgumentsExpressionTests.java
```diff
@@ -42,7 +42,6 @@ void shouldEvaluateExpression(String expression, String expectedResult)
     }
 
     private static class TestExpressionProcessor extends AbstractExpressionProcessor<String>
-            implements NormalizingArguments
```
valfirst, line 45: since no such interface is available anymore, the test class should be renamed to test utility class (or completely removed)

vividus-plugin-datetime/src/main/java/org/vividus/expression/DiffDateExpressionProcessor.java
```diff
@@ -32,11 +32,15 @@
...
 @Named
-public class DiffDateExpressionProcessor extends AbstractExpressionProcessor<String> implements NormalizingArguments
+public class DiffDateExpressionProcessor extends AbstractExpressionProcessor<String>
 {
-    private static final Pattern DIFF_DATE_PATTERN = Pattern
-            .compile("^diffDate\\((.+?),(?<!\\\\,)(.+?),(?<!\\\\,)(.+?),(?<!\\\\,)(.+?)(,(?<!\\\\,)(.+?))?\\)$",
+    // CHECKSTYLE:OFF
+    // @formatter:off
+    private static final Pattern DIFF_DATE_PATTERN = Pattern.compile(
+            "^diffDate\\((\\s*\"\"\".+?\"\"\"|.+?),(?<!\\\\,)(\\s*\"\"\".+?\"\"\"|.+?),(?<!\\\\,)(\\s*\"\"\".+?\"\"\"|.+?),(?<!\\\\,)(\\s*\"\"\".+?\"\"\"|.+?)(,(?<!\\\\,)(.+?))?\\)$",
```
valfirst, line 40: 😵‍💫

valfirst (commented):
Please do not forget to update docs and add integration tests

vividus-plugin-datetime/src/main/java/org/vividus/expression/DiffDateExpressionProcessor.java
EDbarvinsky: https://regex101.com/r/TJWYhc/1

ikalinin1: image

vividus-engine/src/main/java/org/vividus/expression/NormalizingArgumentsUtils.java
```diff
@@ -18,10 +18,25 @@
...
-        return StringUtils.replace(argument.trim(), "\\,", ",");
+    }
+
+    static String normalize(String argument)
+    {
+        if (StringUtils.isNotBlank(argument))
+        {
+            String trimmed = argument.trim();
+            String wrapToken = "\"\"\"";
+            boolean isQuoteWrapped = trimmed.startsWith(wrapToken) && trimmed.endsWith(wrapToken);
```
ikalinin1, line 33: is this check required?

vividus-plugin-datetime/src/main/java/org/vividus/expression/DiffDateExpressionProcessor.java
EDbarvinsky: have not found simpilier solution.

ikalinin1 (approved):

vividus-engine/src/main/java/org/vividus/expression/NormalizingArgumentsUtils.java
```diff
@@ -18,10 +18,26 @@
...
     {
-        return StringUtils.replace(argument.trim(), "\\,", ",");
+    }
+
+    static String normalize(String argument)
+    {
+        if (StringUtils.isNotBlank(argument))
+        {
+            String trimmed = argument.trim();
+            String wrapToken = "\"\"\"";
```
valfirst, line 32: move to constant

vividus-engine/src/main/java/org/vividus/expression/NormalizingArgumentsUtils.java
```diff
@@ -18,10 +18,26 @@
...
+            boolean isQuoteWrapped = trimmed.startsWith(wrapToken) && trimmed.endsWith(wrapToken);
```
valfirst, line 33:
```suggestion
            boolean wrappedInQuotes = trimmed.startsWith(wrapToken) && trimmed.endsWith(wrapToken);
```
is*** - for method names

vividus-plugin-datetime/src/main/java/org/vividus/expression/DiffDateExpressionProcessor.java
```diff
@@ -80,8 +86,9 @@ private String processNegative(Duration duration, String durationAsString)
 
     private ZonedDateTime getZonedDateTime(Matcher expressionMatcher, int inputDateGroup, int inputFormatGroup)
     {
-        DateTimeFormatter inputFormat = DateTimeFormatter
-                .ofPattern(normalize(expressionMatcher.group(inputFormatGroup)), Locale.ENGLISH);
-        return dateUtils.parseDateTime(normalize(expressionMatcher.group(inputDateGroup)), inputFormat);
+        DateTimeFormatter inputFormat = DateTimeFormatter.ofPattern(
+                normalize(expressionMatcher.group(inputFormatGroup)), Locale.ENGLISH);
+        return dateUtils.parseDateTime(normalize(expressionMatcher.group(inputDateGroup)),
+                inputFormat);
```
valfirst, line 92: noisy changes

vividus-plugin-datetime/src/main/java/org/vividus/expression/FormatDateExpressionProcessor.java
```diff
@@ -48,11 +50,11 @@ public FormatDateExpressionProcessor(DateUtils dateUtils)
     @Override
     protected String evaluateExpression(Matcher expressionMatcher)
     {
-        ZonedDateTime zonedDate = dateUtils.parseDateTime(expressionMatcher.group(INPUT_DATE_GROUP),
-                ISO_STANDARD_FORMAT);
+        ZonedDateTime zonedDate = dateUtils.parseDateTime(
+                normalize(expressionMatcher.group(INPUT_DATE_GROUP)), ISO_STANDARD_FORMAT);
```
valfirst, line 54: does it make sense to introduce custom `ExpressionArgumentsMatcher` that will wrap Java matcher and act as a decorator, for example: 
```java
public String getArgument(int argumentIndex)
{
    return normalize(matcher.group(argumentIndex));
}
```
I have not checked whether it fits well for all cases

vividus-plugin-datetime/src/main/java/org/vividus/expression/FormatDateToExpressionProcessor.java
```diff
@@ -46,12 +47,12 @@ public FormatDateToExpressionProcessor(DateUtils dateUtils)
     @Override
     protected String evaluateExpression(Matcher expressionMatcher)
     {
-        String inputDate = normalize(expressionMatcher.group(INPUT_DATE_GROUP));
+        String inputDate = NormalizingArgumentsUtils.normalize(expressionMatcher.group(INPUT_DATE_GROUP));
```
valfirst, line 50: use static method import

vividus-plugin-datetime/src/main/java/org/vividus/expression/ShiftDateExpressionProcessor.java
```diff
@@ -51,10 +53,10 @@ public ShiftDateExpressionProcessor(DateUtils dateUtils)
...
     protected String evaluateExpression(Matcher expressionMatcher)
     {
-        DateTimeFormatter format = DateTimeFormatter.ofPattern(normalize(expressionMatcher.group(FORMAT_GROUP)),
-                DEFAULT_LOCALE);
-        ZonedDateTime zonedDateTime = dateUtils.parseDateTime(normalize(expressionMatcher.group(INPUT_DATE_GROUP)),
-                format);
+        DateTimeFormatter format = DateTimeFormatter.ofPattern(
+                normalize(expressionMatcher.group(FORMAT_GROUP)), DEFAULT_LOCALE);
+        ZonedDateTime zonedDateTime = dateUtils.parseDateTime(
+                normalize(expressionMatcher.group(INPUT_DATE_GROUP)), format);
```
valfirst, line 59: noisy changes

valfirst (changes requested):

vividus-plugin-datetime/src/main/java/org/vividus/expression/FormatDateExpressionProcessor.java
EDbarvinsky: for a nise implementation there is some restriction related to the fact that`Pattern` and `Matcher` are final classes, but lets try to do smth over available interfaces

vividus-engine/src/main/java/org/vividus/expression/ArgumentMatcher.java
```diff
@@ -16,12 +16,9 @@
 
 package org.vividus.expression;
 
-import org.apache.commons.lang3.StringUtils;
-
-interface NormalizingArguments
+public interface ArgumentMatcher
```
valfirst, line 19: no need to use interface here, most likely there won't be multiple implementations

vividus-engine/src/main/java/org/vividus/expression/AbstractExpressionProcessor.java
```diff
@@ -35,11 +35,11 @@ public Optional<T> execute(String expression)
         Matcher expressionMatcher = pattern.matcher(expression);
         if (expressionMatcher.find())
```
uarlouski, line 36: I think it would be more consistent if we initialize ExpressionArgumentMatcher at line 35 and all subsequent calls will be done on the wrapper including find

valfirst: doubtful, we'll need argument matcher only if we find expression

uarlouski: then its better to rename the class because matcher in its name seems too wide, I believe it should be more specific

uarlouski (approved):

vividus/src/main/java/org/vividus/expression/RoundExpressionProcessor.java
```diff
@@ -18,90 +18,59 @@
...
-        public RoundingMode getRoundingMode()
-        {
-            if (roundingMode == null)
-            {
-                return value.charAt(0) == '-' ? RoundingMode.HALF_DOWN : RoundingMode.HALF_UP;
-            }
-            return (RoundingMode) converter.convertValue(roundingMode, RoundingMode.class);
-        }
+        super("round", 1, MAX_ARGS, args -> {
+            String valueToRound = args.get(0);
```
uarlouski, line 50: in general, does it make sense to keep indexes under static variable names? meaning that previously arg 0 was VALUE, arg 1 MAX_FRACTION_DIGITS etc., I understand that variables name that we assign are alear enough, but just to be safe perhaps we could continue name them

uarlouski: discussed

uarlouski (approved):

PR #3487, opened 2022-12-23
[plugin-web-app] Use new headless mode

valfirst:
See details: https://github.com/SeleniumHQ/selenium/pull/11099#issue-1402374967

### DIFF
--- a/vividus-plugin-web-app/src/main/resources/properties/profile/web/headless/chrome/profile.properties
+++ b/vividus-plugin-web-app/src/main/resources/properties/profile/web/headless/chrome/profile.properties
@@ -1,2 +1,2 @@
 selenium.browser=chrome
-web.driver.CHROME.command-line-arguments=--headless --hide-scrollbars
+web.driver.CHROME.command-line-arguments=--headless=chrome --hide-scrollbars


vividus-plugin-web-app/src/main/resources/properties/profile/web/headless/chrome/profile.properties
```diff
@@ -1,2 +1,2 @@
 selenium.browser=chrome
-web.driver.CHROME.command-line-arguments=--headless --hide-scrollbars
+web.driver.CHROME.command-line-arguments=--headless=chrome --hide-scrollbars
```
uarlouski, line 2: can it be a potential breaking change for end users? as per https://chromium.googlesource.com/chromium/src/+/833543a511a970f44fb074b9f30c6b6268e0be71 its completely different browser implementation

valfirst: see initial PR comment

valfirst:
Chrome dev team is still playing with options (https://www.selenium.dev/blog/2023/headless-is-going-away/#what-are-the-two-headless-modes):
> The traditional --headless, and since version 96, Chrome has a new headless mode that allows users to get the full browser functionality (even run extensions). Between versions 96 to 108 it was --headless=chrome, after version 109 --headless=new.

valfirst:
And more details:
https://developer.chrome.com/articles/new-headless/: 
> Currently, passing the `--headless` command-line flag without an explicit value still activates the old Headless mode — but we plan to change this default to new Headless over time.

Unofficial ETA: 2023Q1 - 2023Q2

PR #3523, opened 2023-01-05
[vividus] Add expression to decode Base64 to binary data

ikalinin1:


vividus/src/main/java/org/vividus/expression/DecodeFromBase64ToBinaryData.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.vividus.expression;
+
+import javax.inject.Named;
+
+import org.vividus.util.Base64Utils;
+
+@Named
```
valfirst, line 23: let's start declaring expression beans in `spring.xml`

vividus/src/main/java/org/vividus/util/Base64Utils.java
```diff
@@ -0,0 +1,70 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Base64;
+import java.util.zip.GZIPOutputStream;
+
+public final class Base64Utils
```
valfirst, line 27: there is no much sense to have it, `Base64` from Java is enough concise

valfirst (changes requested):

valfirst (changes requested):

valfirst (approved):

PR #3582, opened 2023-01-27
[engine] Fix parsing of nested expressions without parameters

draker94:
The previous RELUCTANT_EXPRESSION_PATTERN didn't correctly process expressions the number of characters in which was not a multiple of 3

### DIFF
--- a/vividus-engine/src/test/java/org/vividus/steps/ExpressionAdapterTests.java
+++ b/vividus-engine/src/test/java/org/vividus/steps/ExpressionAdapterTests.java
@@ -44,6 +44,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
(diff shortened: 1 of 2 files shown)


vividus-engine/src/test/java/org/vividus/steps/ExpressionAdapterTests.java
```diff
@@ -116,14 +116,18 @@ void testSupportedExpressionNestedExpr()
     }
 
     @SuppressWarnings("unchecked")
-    @Test
-    void testNestedExpressionWithoutParams()
+    @ParameterizedTest
+    @CsvSource({
```
valfirst, line 120: use `@ValueSource` for list of values (`@CsvSource` is used for 2-dimensions data aka tables)

vividus-engine/src/test/java/org/vividus/steps/ExpressionAdapterTests.java
```diff
@@ -116,14 +117,15 @@ void testSupportedExpressionNestedExpr()
     }
 
     @SuppressWarnings("unchecked")
-    @Test
-    void testNestedExpressionWithoutParams()
+    @ParameterizedTest
+    @ValueSource(strings = { EXPRESSION_KEYWORD, "targets" })
```
ikalinin1, line 121: could you please elaborate how these changes cover fix?

ikalinin1 (approved):

vividus-engine/src/test/java/org/vividus/steps/ExpressionAdapterTests.java
draker94: The previous regex correctly matches nested expressions without arguments only if the number of characters in this expression is multiple of three (3+3+3...). For example, it correctly parses "story_random" or "target", but not "story_indexed" or "targets". This test checks the parsing of expressions with different number of characters in order to eliminate similar problems in the future (this is quite possible, considering regex complexity)

PR #3599, opened 2023-01-30
Mobile Plugin

daryashev:


ikalinin1:
@daryashev  can you please attach the generated docs as an archive?

docs/modules/plugins/partials/mobile-ios-simulator.adoc
```diff
@@ -0,0 +1,25 @@
+====
+
+. Open project_ios repo in code management system
```
ikalinin1, line 3: Looks like too specific instructions, just clone should be enough

docs/modules/plugins/partials/mobile-ios-simulator.adoc
```diff
@@ -0,0 +1,25 @@
+====
```
ikalinin1, line 1: What is the purpose of these instructions?

docs/modules/plugins/partials/mobile-installation.adoc
```diff
@@ -0,0 +1,34 @@
+
+=== Setup environment for Mobile testing
```
ikalinin1, line 2: Can we modify - reorganize appium-setup instead?

docs/modules/plugins/partials/mobile-android-emulator.adoc
```diff
@@ -0,0 +1,19 @@
+====
+
+. Open Android Studio
```
ikalinin1, line 3: I think these docs cover it? https://developer.android.com/studio/run/managing-avds

PR #3601, opened 2023-02-01
[plugin-web-app] Record failed assertion for intercepted click

draker94:


docs/modules/plugins/partials/plugin-web-app-steps.adoc
```diff
@@ -98,6 +98,11 @@ The atomic actions performed are:
 words if https://www.selenium.dev/exceptions/#stale_element_reference[StaleElementReferenceException]
 is occurred at any atomic action.
 
+[NOTE]
+====
+If element by the specified locator is not clickable (overlapped by another element, problems with page/context loading or the element is disabled) the step will fail with corresponding error.
```
valfirst, line 103: would you mind adding error example here?

valfirst (changes requested):

valfirst (approved):

uarlouski (approved):

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/MouseActions.java
```diff
@@ -91,10 +91,9 @@ private void tryToWorkaroundException(WebElement element, Optional<Action> defau
         {
             try
             {
-                if (webDriverManager.isBrowserAnyOf(Browser.CHROME)
-                        && message.contains(". Other element would receive the click"))
+                if (message.contains(". Other element would receive the click"))
                 {
-                    javascriptActions.click(element);
+                    softAssert.recordFailedAssertion(COULD_NOT_CLICK_ERROR_MESSAGE + webDriverException);
```
ikalinin1, line 96: Why are you appending an exception? What is the resulting error?
org.vividus.softassert.SoftAssert#recordFailedAssertion(java.lang.String, java.lang.Throwable)

ikalinin1 (changes requested):

vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/MouseActions.java
draker94: > What is the resulting error?
interException
> Why are you appending an exception?
It has pretty enough information about the error
> org.vividus.softassert.SoftAssert#recordFailedAssertion(java.lang.String, java.lang.Throwable)
I think previous way (see previous pic) in the result a little readable
image

ikalinin1: IMHO you should record the exception with stacktrace

valfirst: for this case that's doubtful, stack trace won't give any useful information

draker94: In any case, we can expand error in Allure for stack trace
image

ikalinin1: Since you all agree on the matter, it is up to you how to deal with the stacktrace. However, that doesn't explain why you are appending an exception instead of its message

ikalinin1: Yep, is it useful for the user to have the class name in error?

PR #3620, opened 2023-02-07
[docs] Document default values in variables

draker94:


docs/modules/commons/pages/variables.adoc
```diff
@@ -160,3 +160,19 @@ Then `${var}` is equal to `story`
 ----
 
 `var` variable will be resolved to value `scenario` at the validation point in the first scenario, but it will be resolved to value `story` in the second scenario.
+
+=== Default values
+
+If there is a possibility that a variable will not be initialized at the current point, it is possible to specify a default value for it using the operator `:` (colon).
+
+.Open either user specified page or default page
```
ikalinin1, line 168: can you replace it with a table with examples of different cases.
see
https://github.com/vividus-framework/vividus/blob/master/vividus-engine/src/test/java/org/vividus/variable/VariablesTests.java#L53-L83

also i think makes sense to mention how will work ${variable:} case

draker94: > can you replace it with a table with examples of different cases

Do you mean "examples of different cases" just for default values or globally for any cases with different variables?
And do we really have to replace the selected fragment with a table?
I think it would be useful to have a short explanation + a real example of use

ikalinin1: I have on my mind a table  like this:

|reference|value|resolved to|
|-------|------|------------|
|${var:123}| value|value|
|${var:123}| not set|123|

etc

ikalinin1: I have on my mind a table  like this:

|reference|value|resolved to|
|-------|------|------------|
|${var:123}| value|value|
|${var:123}| not set|123|

etc

docs/modules/commons/pages/variables.adoc
```diff
@@ -160,3 +160,62 @@ Then `${var}` is equal to `story`
...
+
+=== Default values
+
+Variable reference can be set to default value using the operator `:` (colon). This value will be used if the variable is uninitialized.
+
+.Examples of using default values
+[cols="1,2,2", options="header"]
+|===
+|reference
+|reference value
```
ikalinin1, line 172:
```suggestion
|variable value
```

ikalinin1 (approved):

docs/modules/commons/pages/variables.adoc
```diff
@@ -160,3 +160,62 @@ Then `${var}` is equal to `story`
...
 `var` variable will be resolved to value `scenario` at the validation point in the first scenario, but it will be resolved to value `story` in the second scenario.
+
+=== Default values
+
+Variable reference can be set to default value using the operator `:` (colon). This value will be used if the variable is uninitialized.
+
+.Examples of using default values
+[cols="1,2,2", options="header"]
+|===
+|reference
```
valfirst, line 171:
```suggestion
|Reference to resolve
```

docs/modules/commons/pages/variables.adoc
```diff
@@ -160,3 +160,62 @@ Then `${var}` is equal to `story`
...
+|variable value
+|output
```
valfirst, line 173:
```suggestion
|Output
```

valfirst (changes requested):

valfirst (approved):

PR #3622, opened 2023-02-08
[tests] Fix flaky integration test for file upload

valfirst:


vividus-tests/src/main/resources/story/integration/ElementStepsTests.story
```diff
@@ -66,6 +66,8 @@ Then the context element has the CSS property 'color' containing '(0, 0, 0, 1)'
 
 Scenario: Step verification When I select an element '$locator' and upload the file '$filePath'
 Given I am on page with URL `http://demo.guru99.com/test/upload/`
+When I switch to frame located `By.id(gdpr-consent-notice)`
```
ikalinin1, line 69: can it be done without UI?

ikalinin1: Do we need to switch back to the page here?

valfirst: > can it be done without UI?

in this case it's better to use UI interactions: 
1) binding to cookie names may lead to failures in future in case of some changes, then it'll be hard to investigate since the system is 3-rd party for us
2) it will still require page refresh, so no performance profit here

valfirst: > Do we need to switch back to the page here?

nope, it's done automactially

uarlouski (approved):

PR #3627, opened 2023-02-10
[docs-plugins] Fix MSSQL connection examples

VitaliLebedzeu:


docs/modules/plugins/pages/plugin-db.adoc
```diff
@@ -109,15 +109,15 @@ implementation(group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '1
 .environment.properties
 [source,properties]
 ----
-db.connection.vividus.url=jdbc:sqlserver://mssql.db.server:1433\mssql_instance;databaseName=my_database
+db.connection.vividus.url=jdbc:sqlserver://;serverName=mssql.db.server;databaseName=my_database;encrypt=true;trustServerCertificate=true
```
valfirst, line 112: Why are 2 last parameters needed? In general we try to keep such examples as minimal as possible

docs/modules/plugins/pages/plugin-db.adoc
```diff
@@ -109,15 +109,15 @@ implementation(group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '1
...
 db.connection.vividus.username=vividus
 db.connection.vividus.password=pa$$w0rd
 ----
 
 .https://docs.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url[URL template]
 [source,properties]
 ----
-jdbc:sqlserver://serverName\instanceName:portNumber;parameter1=value1;parameter2=value2
+jdbc:sqlserver://;servername=server_name;encrypt=true;integratedSecurity=true;parameter1=value1;parameter2=value2
```
valfirst, line 120: Why are `encrypt` and `integratedSecurity` parameters needed?

valfirst (changes requested):

docs/modules/plugins/pages/plugin-db.adoc
VitaliLebedzeu: for the secure connection

VitaliLebedzeu: from the MS documentation
image

docs/modules/plugins/pages/plugin-db.adoc
VitaliLebedzeu: From the MS documentation
- the **encrypt=true** essential if we connect to the local DB
image

- the **encrypt=true** and **trustServerSertificate=true** essential if we connect to the remote DB
image

valfirst: 1) What's the reason to change the original format around server? Is old one wrong? Or does new one have benefits?
2) `encrypt` - is not mandatory parameter and it has many peculiarities (read the docs), so I prefer to do not add it without any strong reason
3) `trustServerCertificate` works only with `encrypt`

docs/modules/plugins/pages/plugin-db.adoc
valfirst: `integratedSecurity` is usually for Windows or require another parameters

valfirst:
@VitaliLebedzeu please check fixes in #3640

VitaliLebedzeu:
@valfirst it looks good. In case of keeping such examples as minimal as possible, I've closed this PR.

PR #3641, opened 2023-02-15
[vividus] Unify naming of classes used as entry points for Gradle tasks

valfirst:


vividus/src/main/java/org/vividus/runner/ScenariosCounter.java
```diff
@@ -0,0 +1,116 @@
...
+        System.out.printf(FORMATTER, number, message);
+        System.out.println();
+    }
+
+    private static void configureStoryLocation(String storyLocation)
+    {
+        Properties properties = BeanFactory.getBean("properties", Properties.class);
+        properties.put("bdd.story-loader.batch1.resource-location", storyLocation);
+        properties.put("bdd.story-loader.batch1.resource-include-patterns", "**/*.story");
+        properties.put("bdd.story-loader.batch1.resource-exclude-patterns", "");
```
uarlouski, line 97: should not it be changed as well?

ikalinin1 (approved):

vividus/src/main/java/org/vividus/runner/ScenariosCounter.java
valfirst: nope, not in this PR, the goal of this PR is to update naming of classes, and then switch to use new classes in build system 2.0

PR #3687, opened 2023-02-28
[plugin-kafka] Add ability to send Kafka event with specifying a key

### LINKED ISSUE #2915: Add possibility to send message in kafka topic with key
I'm and team members are always frustrated when we can only send kafka message using the step 'When I send data `$data` to `$producerKey` Kafka topic `$topic`' without key mentioned there. 
As the key is used by kafka to put data in a specific partition which is how kafka allows for large streams of data.
So developers asked to add possibility to send kafka message with key using vividus to process data correctly when auto tests are running.

Add step that allows to send kafka message WITH key

**Describe alternatives you've considered**
Unfortunately I can't suggest alternative here :( 
Maybe we can specify partition instead of key, but I think that it's better to specify the key anyway 

**Additional context**
-

uarlouski:
Closes #2915

### DIFF
--- a/vividus-plugin-kafka/src/main/java/org/vividus/steps/kafka/KafkaSteps.java
+++ b/vividus-plugin-kafka/src/main/java/org/vividus/steps/kafka/KafkaSteps.java
@@ -62,6 +62,8 @@ import org.vividus.util.property.IPropertyParser;
 import org.vividus.util.wait.DurationBasedWaiter;
 import org.vividus.variable.VariableScope;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
 public class KafkaSteps
 {
     private static final String DOT = ".";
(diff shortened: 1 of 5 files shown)


vividus-plugin-kafka/src/main/java/org/vividus/steps/kafka/KafkaSteps.java
```diff
@@ -113,19 +115,42 @@ private <T> Map<String, T> convert(String propertiesPrefix, IPropertyParser prop
     }
 
     /**
-     * Send the event with the specified value to the provided topic with no key or partition.
+     * Sends the event with the key and the value to the provided topic. Events with the same event key are written to
```
valfirst, line 118: it doesn't make sense to put info about event keys to step which doesn't accept key

valfirst (changes requested):

vividus-plugin-kafka/src/main/java/org/vividus/steps/kafka/KafkaSteps.java
uarlouski: ops, it had to be the step that accepts key :)

PR #3778, opened 2023-03-27
[soft-assert] Fix missing attachments with enabled fail-fast at scenario level

draker94:


vividus-soft-assert/src/main/java/org/vividus/softassert/ISoftAssert.java
```diff
@@ -31,6 +32,16 @@
      */
     boolean assertTrue(String description, boolean condition);
 
+    /**
+     * Asserts that a condition is true. If it isn't, an assertion error is added to the collection.
+     * Can be specified an additional operation which will always be executed.
```
valfirst, line 37:
```suggestion
     * An additional operation which will always be executed is specified.
```

valfirst: also please add the description the operation is extremely useful when fail-fast is enabled

vividus/src/main/java/org/vividus/steps/ArchiveSteps.java
```diff
@@ -110,28 +112,41 @@ public void saveArchiveEntriesToVariables(DataWrapper archiveData, List<ArchiveV
...
-                    comparisonRule, expectedName), entryNames, hasItem(comparisonRule.createMatcher(expectedName)));
+                    comparisonRule, expectedName), entryNames, hasItem(comparisonRule.createMatcher(expectedName)),
+                    publishAttachmentOnFailure(entryNames, reportNotPublished));
         }
-        return softAssert.assertThat("The archive contains entry with name " + expectedName, entryNames,
-                hasItem(expectedName));
+        else
+        {
+            softAssert.assertThat("The archive contains entry with name " + expectedName, entryNames,
+                    hasItem(expectedName), publishAttachmentOnFailure(entryNames, reportNotPublished));
```
valfirst, line 135: it makes sense to refactor to have only one invocation of `softAssert`

vividus-soft-assert/src/main/java/org/vividus/softassert/ISoftAssert.java
```diff
@@ -31,6 +32,16 @@
...
+     * @param description The assertion description
+     * @param condition Condition to be checked
+     * @param resultConsumer Always executed additional operation that consumes result of the assertion
```
valfirst, line 40: it seems it's used only when assertion is failed, does it make sense to convert it to something like: `Runnable failureHandler`?

valfirst (changes requested):

vividus-soft-assert/src/main/java/org/vividus/softassert/ISoftAssert.java
draker94: We have two places where we publish attachments unconditionally: [ResourceCheckSteps](https://github.com/vividus-framework/vividus/blob/66ff1945a8001219a735e61a39715a01178e414f/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java#L128-L129) and [HttpResourceValidationSteps](https://github.com/vividus-framework/vividus/blob/6e1be1484f06d59151f7de0aadbe72ace2e5c20f/vividus-plugin-rest-api/src/main/java/org/vividus/steps/http/HttpResourceValidationSteps.java#L63-L64). These cases will be fixed in separate PR.

valfirst (approved):

ikalinin1 (approved):

EDbarvinsky (approved):

uarlouski (approved):

PR #3800, opened 2023-04-04
[plugin-web-app] Introduce OS independent `CONTROL` key

valfirst:


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/WebAtomicActionFactories.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+    }
+
+    private static void buildKeysActions(KeysManager keysManager, List<String> keys,
+            Consumer<CharSequence> actionBuilder)
+    {
+        notEmpty(keys, "At least one key should be provided");
+        for (String key : keys)
+        {
+            actionBuilder.accept(key.length() == 1 ? key : keysManager.convertToKey(true, key));
```
uarlouski, line 43: why don't we perform key validation in case of one key?

valfirst: this logic was here before: https://github.com/vividus-framework/vividus/blob/c01f672b1e6fdf12df6b7c335dfb9a769c2fa36f/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/WebSequenceActionType.java#L198-L213

uarlouski (approved):

PR #3825, opened 2023-04-12
[plugin-web-app] Add step to close browser

uarlouski:


docs/modules/plugins/pages/plugin-web-app.adoc
```diff
@@ -166,6 +166,15 @@ include::plugins:partial$plugin-web-app-profiles.adoc[]
 
 include::plugins:partial$plugin-web-app-steps.adoc[]
 
+=== Close browser
+
+Closes the browser and quits the remote session.
```
valfirst, line 171: What if user runs tests locally (remote session might be confusing)? I believe there is no sense to put this info

vividus-tests/src/main/resources/story/integration/BrowserSteps.story
```diff
@@ -0,0 +1,13 @@
+Meta:
+    @epic vividus-plugin-web-app
+
+Scenario: Step verification "When I close browser"
+Given I am on main application page
+Given I initialize scenario variable `item-key` with value `mark#{generate(regexify '[a-z]{10}')}`
+Given I initialize scenario variable `item-value` with value `mark#{generate(regexify '[a-z]{10}')}`
+When I execute javascript `localStorage.setItem('${item-key}', '${item-value}'); return localStorage.getItem('${item-key}');` and save result to scenario variable `set-value`
```
valfirst, line 8: What about using dedicated steps?

valfirst (changes requested):

valfirst (approved):

PR #3834, opened 2023-04-14
[aws] Add step to configure AWS authentication

### LINKED ISSUE #3806: Add possibility to use more than one AWS S3 account in one test run
We need to test S3 bucket objects from two different AWS accounts. The goal of this test is to verify visibility/restricted access of S3 bucket objects for second AWS account.

We have one test scenario, where we start to work with S3 bucket as one user (first account, credentials are specified through properties: system.aws.accessKeyId and system.aws.secretKey). Under this user we perform step:
``When I upload data `$data` with key `$objectKey` and content type `$contentType` to S3 bucket `$bucketName``
than we request through API credentials for second account (saving AWS_ACCESS_KEY_ID и AWS_SECRET_ACCESS_KEY from API response) 
and need to perform under second user next steps:
```
When I fetch object with key `$objectKey` from S3 bucket `$bucketName` and save result to $scopes variable `$variableName`
When I delete object with key `$objectKey` from S3 bucket `$bucketName`
When I upload data `$data` with key `$objectKey` and content type `$contentType` to S3 bucket `$bucketName`
```

valfirst:
Closes #3806


vividus-extension-aws/build.gradle
```diff
@@ -0,0 +1,14 @@
+project.description = 'VIVIDUS extension for AWS Cloud'
+
+dependencies {
+    api platform(group: 'com.amazonaws', name: 'aws-java-sdk-bom', version: '1.12.445')
```
ikalinin1, line 4: does it make sense to delete this dependency in dependent modules?

ikalinin1 (approved):

uarlouski (approved):

PR #3850, opened 2023-04-18
[xray-exporter] Add ability to globally disable test case updates

uarlouski:


vividus-to-xray-exporter/src/main/java/org/vividus/xray/facade/XrayFacade.java
```diff
@@ -82,6 +84,14 @@ public <T extends AbstractTestCase> String createTestCase(T testCase) throws IOE
     public <T extends AbstractTestCase> void updateTestCase(String testCaseKey, T testCase)
             throws IOException, NonEditableIssueStatusException, JiraConfigurationException
     {
+        if (!testCaseUpdatesEnabled)
```
valfirst, line 87: I think this check should be done upper (in the method invoking this one)

valfirst (approved):

ikalinin1 (approved):

PR #3854, opened 2023-04-19
[xray-exporter] Avoid creating of 'tests' link if it already exists

uarlouski:


vividus-to-xray-exporter/src/main/java/org/vividus/xray/facade/XrayFacade.java
```diff
@@ -147,6 +148,21 @@ private void checkIfIssueEditable(String issueKey)
     public void createTestsLink(String testCaseId, String requirementId) throws IOException, JiraConfigurationException
     {
         String linkType = "Tests";
+        JiraEntity issue = jiraFacade.getIssue(testCaseId);
```
valfirst, line 151: is this call needed? don't we have testCase instance?

uarlouski: discussed

valfirst (approved):

ikalinin1 (approved):

PR #3871, opened 2023-04-21
[plugin-rest-api] Split resource validator for different types of validation

### LINKED ISSUE #3849: `ClassCastException` when using different resource validators
**Describe the bug**
If try to use different resource validation steps in single test run we get
`java.util.concurrent.ExecutionException: java.lang.ClassCastException`

**To Reproduce**
Use the next steps together in single test run:
`Then HTTP resources are valid:$resources`
``Then all resources by selector `$cssSelector` are valid on:$pages``

**Expected behavior**
Resource validation steps have to work together.

**Artifacts**
[allureResourceCheckerException.zip](https://github.com/vividus-framework/vividus/files/11258840/allureResourceCheckerException.zip)

draker94:
Closes #3849

### DIFF
--- a/vividus-plugin-rest-api/src/main/resources/vividus-plugin/spring.xml
+++ b/vividus-plugin-rest-api/src/main/resources/vividus-plugin/spring.xml
@@ -7,21 +7,17 @@
         http://www.springframework.org/schema/util https://www.springframework.org/schema/util/spring-util.xsd"
     default-lazy-init="true">
 
-    <bean class="org.vividus.http.validation.ResourceValidator" >
-      <constructor-arg index="0">
-        <bean class="org.vividus.http.client.HttpClient"
-            factory-bean="httpClientFactory"
-            factory-method="buildHttpClient">
-            <constructor-arg>
-                <bean parent="restApiHttpClientConfig">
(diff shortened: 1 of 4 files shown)


vividus-plugin-rest-api/src/main/resources/vividus-plugin/spring.xml
```diff
@@ -7,7 +7,7 @@
         http://www.springframework.org/schema/util https://www.springframework.org/schema/util/spring-util.xsd"
     default-lazy-init="true">
 
-    <bean class="org.vividus.http.validation.ResourceValidator" >
+    <bean class="org.vividus.http.validation.ResourceValidator" scope="prototype">
```
valfirst, line 10: that's the easiest solution, but in future it may lead to issue if one resource validator is needed to be injected in 2 places, so it's better:
- to declare a separate bean of HTTP client for ResourceValidator: e.g. `resourceValidatorHttpClient`
- inject 2 beans of `ResourceValidator` in 2 different places (and reuse HTTP client bean)

valfirst (approved):

uarlouski (approved):

PR #3898, opened 2023-04-28
[plugin-web-app-to-rest-api] Filter out from validation img with data…

uarlouski:
… URLs

### DIFF
--- a/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java
+++ b/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java
@@ -157,6 +157,11 @@ public class ResourceCheckSteps
     private Optional<WebPageResourceValidation> parseElement(Element element)
     {
         String elementUriAsString = getElementUri(element).trim();
+        if (elementUriAsString.startsWith("data:"))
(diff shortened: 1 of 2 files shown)


vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java
```diff
@@ -157,6 +157,11 @@ private Stream<WebPageResourceValidation> createResourceValidations(Collection<E
     private Optional<WebPageResourceValidation> parseElement(Element element)
     {
         String elementUriAsString = getElementUri(element).trim();
+        if ("img".equals(element.tagName()) && elementUriAsString.startsWith("data:"))
```
valfirst, line 160: shouldn't we filtered out all data URLs?

valfirst (changes requested):

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java
uarlouski: as per examples in the integer its used only in images

valfirst (approved):

PR #3945, opened 2023-05-16
[tests] Rely on viewport size calculated by selenium

uarlouski:


valfirst (approved):

valfirst:
- is viewport size really calculated by Selenium?
- do we need to deprecate the property?

uarlouski:
> * is viewport size really calculated by Selenium?

bad commit message, it's calculated by us here https://github.com/vividus-framework/vividus/blob/master/vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/WebJavascriptActions.java#L242

> * do we need to deprecate the property?

I think it makes sense to see how it goes with the calculated viewport over chrome version updates

PR #3951, opened 2023-05-18
[docs] Add docs for web application context steps

draker94:


docs/modules/plugins/partials/plugin-web-app-steps.adoc
```diff
@@ -622,7 +623,7 @@ When I wait until browser console infos by regex `.*Application ready.*` appear
 Then `${logs}` matches `.*Application ready in \d+ seconds.*`
 ----
 
-=== Perform steps for each found element
+==== Perform steps for each found element
```
valfirst, line 626: in this case 3-level header is missing or incorrect

docs/modules/plugins/partials/plugin-web-app-steps.adoc
```diff
@@ -654,3 +655,77 @@ When I find = `5` elements by `By.xpath(//script):a` and for each element do
...
+
+.Switch to frame
+----
+Given I am on page with URL `https://vividus-test-site.onrender.com/nestedFrames.html`
+Then text `Modal Frame Example` does not exist
+When I switch to frame located `id(parent)`
+Then text `Modal Frame Example` exists
+----
+
+==== Switch context to new window
```
valfirst, line 695: let's avoid `window` term (it's too technical and not clear for users) and use `tab` instead

valfirst (changes requested):

docs/modules/plugins/partials/plugin-web-app-steps.adoc
draker94: > let's avoid `window` term (it's too technical and not clear for users) and use `tab` instead

Okay, but since we mention the term `window` in step wording and params, I think it makes sense to give a little explanation of it

valfirst: all "window"-based syntax steps will be updated to use "tab"

valfirst (approved):

docs/modules/plugins/partials/plugin-web-app-steps.adoc
```diff
@@ -654,3 +654,78 @@ When I find = `5` elements by `By.xpath(//script):a` and for each element do
...
+Switch the focus of future browser commands to new tab with specified title.
+
+[source,gherkin]
+----
+When I switch to window with title that $stringComparisonRule `$windowName`
+----
+* `$stringComparisonRule` - xref:parameters:string-comparison-rule.adoc[String comparison rule].
+* `$windowName` - The expected tab title.
+
+.Open the new tab by link and switch to it using regex title pattert
```
valfirst, line 724:
```suggestion
.Open the new tab by link and switch to it using regex title pattern
```

PR #3952, opened 2023-05-18
[vividus] Add ability to filter rows by range of indices in FILTERING transformer

IBrun:


vividus/src/main/java/org/vividus/transformer/FilteringTableTransformer.java
```diff
@@ -129,7 +129,19 @@ else if (byRandomRows != null)
         else
         {
             return Stream.of(StringUtils.split(byRowIndexes, ';'))
-                    .mapToInt(Integer::parseInt)
+                    .flatMapToInt(range ->
+                    {
+                        String[] rangeBounds = StringUtils.split(range, "-", 2);
+                        int startInclusive = Integer.parseInt(rangeBounds[0]);
```
valfirst, line 135:
```suggestion
                        int startInclusive = Integer.parseInt(rangeBounds[0].trim());
```
?

vividus/src/main/java/org/vividus/transformer/FilteringTableTransformer.java
```diff
@@ -129,7 +129,19 @@ else if (byRandomRows != null)
...
+                        if (rangeBounds.length == 1)
+                        {
+                            return IntStream.of(startInclusive);
+                        }
+                        int endInclusive = Integer.parseInt(rangeBounds[1]);
```
valfirst, line 140:
```suggestion
                        int endInclusive = Integer.parseInt(rangeBounds[1].trim());
```
?

valfirst (changes requested):

vividus/src/main/java/org/vividus/transformer/FilteringTableTransformer.java
```diff
@@ -129,7 +129,19 @@ else if (byRandomRows != null)
...
+                        isTrue(startInclusive <= endInclusive, "'byRowIndexes' has invalid range of indexes: "
```
valfirst, line 141: need to use the same naming across docs and code: either `indices` or `indexes`

valfirst (approved):

PR #3976, opened 2023-05-24
[plugin-accessibility] Optimize locator conversions

uarlouski:


vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
```diff
@@ -86,7 +86,6 @@ public void checkAccessibility(List<AccessibilityCheckOptions> checkOptions)
     {
         URI currentUrl = UriUtils.createUri(webDriverProvider.get().getCurrentUrl());
         checkOptions.stream()
-                    .filter(o -> null == o.getElementsToCheck() || !o.getElementsToCheck().isEmpty())
```
valfirst, line 89: why is it removed?

uarlouski: - I believe this verification is useless because as per code logic elementsToCheck can be null or not empty string here, we didn't even log anything. 
- after optimisation, ElementsToCheck is a collection (empty or not without nulls) so all accessibility tests where ElementsToCheck is not specified will not be executed silently

uarlouski: good catch, fixed

valfirst (approved):

PR #4000, opened 2023-05-31
[engine] Automate logging of warnings on deprecated composite steps

draker94:


vividus-extension-selenium/src/main/resources/steps/defaults/deprecated.steps
```diff
@@ -1,17 +1,17 @@
 Composite: When I wait until element located `$locator` appears
-!-- WARNING: The step "When I wait until element located `$locator` appears" is deprecated and will be removed in VIVIDUS 0.6.0
+!-- VersionToRemoveStep: "0.6.0" ReplacementFormatPattern: "When I wait until element located by `%1$s` appears"
```
uarlouski, line 2: RemovalVersion: ReplacementPattern:
isn't it more friendly?

draker94: > RemovalVersion: ReplacementPattern: isn't it more friendly?

Yep
But I tried to match the values of the `Replacement` annotation to keep everything consistent
Need to discuss this moment

vividus-engine/src/main/java/org/vividus/StepReplacementStoryReporter.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jbehave.core.model.Step;
+import org.jbehave.core.steps.StepCreator.StepExecutionType;
+import org.vividus.annotation.ReplacementMonitor;
+
+public class StepReplacementStoryReporter extends ChainedStoryReporter
+{
+    private static final Pattern DEPRECATED_COMPOSITE_STEP_COMMENT_PATTERN = Pattern
+            .compile("^!--\\s+VersionToRemoveStep:\\s+\"(.*)\"\\s+ReplacementFormatPattern:\\s+\"(.*)\"$");
+
+    private final ThreadLocal<String> currentDeprecatedStepLogMessage = ThreadLocal
```
valfirst, line 32: it's better to use `TestContext`

vividus-engine/src/main/java/org/vividus/annotation/ReplacementMonitor.java
```diff
@@ -61,22 +61,25 @@ public void stepMatchesPattern(String step, boolean matches, StepPattern pattern
...
-            Matcher stepMatcher = getStepMatcher(rawDeprecatedStep);
-            String[] stepParams = getParametersFromStep(stepMatcher);
+        getAnnotation(method).ifPresent(r -> LOGGER.info(
+                getLogMessageForRunningDeprecatedStep(r.versionToRemoveStep(), r.replacementFormatPattern())));
+    }
 
-            try (Formatter newStep = new Formatter())
-            {
-                newStep.format(r.replacementFormatPattern(), stepParams);
+    public String getLogMessageForRunningDeprecatedStep(String versionToRemove, String replacementFormatPattern)
```
valfirst, line 68: it should be moved to some other class, and then it should be reused in both StoryReporter and StepMonitor

vividus-extension-selenium/src/main/resources/steps/defaults/deprecated.steps
valfirst, line 2:
```suggestion
!-- Deprecated: 0.6.0, When I wait until element located by `%1$s` appears
```
we can minimize the number of required text

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/annotation/ReplacementMonitor.java
draker94: > it should be moved to some other class, and then it should be reused in both StoryReporter and StepMonitor

We need gathered data from this monitor to get deprecation messaged
Probably it would better to left this logic here, but move logging part to separate monitor

vividus-engine/src/main/java/org/vividus/replacement/CompositeStepsReplacementStoryReporter.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public void comment(String step)
+    {
+        String comment = step;
+        String deprecatedStepLogMessage = testContext.remove(DEPRECATED_STEP_LOG_MESSAGE_KEY);
+        if (deprecatedStepLogMessage != null)
+        {
+            comment = deprecatedStepLogMessage;
+        }
+        super.comment(comment);
```
uarlouski, line 70: simplify code using ternary operator?

vividus-engine/src/main/java/org/vividus/replacement/LogReplacementStepMonitor.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public void beforePerforming(String step, boolean dryRun, Method method)
+    {
+        getReplacementAnnotation(method).ifPresent(r -> LOGGER.info(replacementStepMonitor
+                .getLogMessageForRunningDeprecatedStep(r.versionToRemoveStep(), r.replacementFormatPattern())));
+    }
+
+    private Optional<Replacement> getReplacementAnnotation(Method method)
```
uarlouski, line 45: can be pretty inlined

vividus-engine/src/main/java/org/vividus/replacement/CompositeStepsReplacementStoryReporter.java
```diff
@@ -0,0 +1,72 @@
...
+    private static final Object DEPRECATED_STEP_LOG_MESSAGE_KEY = CompositeStepsReplacementStoryReporter.class;
+
+    private final TestContext testContext;
+    private final ReplacementStepMonitor replacementStepMonitor;
+
+    public CompositeStepsReplacementStoryReporter(TestContext testContext,
+            ReplacementStepMonitor replacementStepMonitor)
+    {
+        this.testContext = testContext;
+        this.replacementStepMonitor = replacementStepMonitor;
```
uarlouski, line 40: not sure if its good when one lifecycle reporting class depends on another lifecycle reporting class

draker94: > not sure if its good when one lifecycle reporting class depends on another lifecycle reporting class

What's the problem in this case?
The sequence of method calls in this case is quite clear

uarlouski: I would say that we add one more responsibility to monitor class that goes beyond monitoring, so its like mixing different things up

valfirst: just move `getLogMessageForRunningDeprecatedStep` to a new separate class and reuse it in both places

vividus-engine/src/main/java/org/vividus/replacement/DeprecatedStepsReplacementProvider.java
```diff
@@ -0,0 +1,85 @@
...
+    private final CollectingStepPatternsMonitor collectingStepPatternsMonitor;
+
+    public DeprecatedStepsReplacementProvider(RunTestContext runTestContext,
+            CollectingStepPatternsMonitor collectingStepPatternsMonitor)
+    {
+        this.runTestContext = runTestContext;
+        this.collectingStepPatternsMonitor = collectingStepPatternsMonitor;
+    }
+
+    protected String getLogMessageForRunningDeprecatedStep(String versionToRemove, String replacementFormatPattern)
```
valfirst, line 42:
```suggestion
    protected String createDeprecatedStepNotification(String versionToRemove, String replacementFormatPattern)
```

vividus-engine/src/main/java/org/vividus/replacement/CompositeStepsReplacementStoryReporter.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jbehave.core.model.Step;
+import org.jbehave.core.steps.StepCreator.StepExecutionType;
+import org.vividus.ChainedStoryReporter;
+import org.vividus.testcontext.TestContext;
+
+public class CompositeStepsReplacementStoryReporter extends ChainedStoryReporter
```
valfirst, line 27:
```suggestion
public class DeprecatedCompositeStepsReporter extends ChainedStoryReporter
```

valfirst (approved):

uarlouski (approved):

PR #4001, opened 2023-05-31
[plugin-accessibility] Change accessibility step naming

uarlouski:


vividus-build-system
valfirst, line 1: redundant changes

valfirst (changes requested):

valfirst (approved):

PR #4017, opened 2023-06-06
[allure-adaptor] Exclude Allure GA plugin

valfirst:


uarlouski:
isn't is better to use ALLURE_NO_ANALYTICS var for this?

uarlouski:
> > isn't is better to use ALLURE_NO_ANALYTICS var for this?
> 
> but how to force it for each user?

I forgot we can't set env vars :)

uarlouski (approved):

PR #4019, opened 2023-06-06
[docs] Improve docs for step filtering blobs

### LINKED ISSUE #3977: plugin-azure-storage-account: misleading documentation
Hi, team!)

I think documentation for plugin-azure-storage-account could be more specific
Ref: https://github.com/vividus-framework/vividus/blob/13606bbb271d71f0a34a9136bd64171f2a160c15/docs/modules/plugins/pages/plugin-azure-storage-account.adoc?plain=1#LL242C4-L242C4

According to the documentation "The full or partial blob name to be matched". I got it like I can include part of the blob name avoiding `blobNamePrefix` repetition when using `matches` rule as `blobNameFilterRule`.
But it turned out that VIVIDUS is applying given regexp to full blob name including part from `blobNamePrefix`.

**So I suggest to change wording to something like**
The full or partial blob name to be matched. It should be specified along with blobNameFilterRule. Regular expression applied with `matches` `blobNameFilterRule` will be applied to the whole blob name, including prefix specified in `blobNamePrefix`.

Or maybe create something like Note window like this
image

Thanks for the great tool btw! Helps me a lot!

valfirst:
Closes #3977

### DIFF
--- a/docs/modules/plugins/pages/plugin-azure-storage-account.adoc
+++ b/docs/modules/plugins/pages/plugin-azure-storage-account.adoc
@@ -239,8 +239,8 @@ It should be specified along with `blobNameFilterValue`.
 |`MATCHES`
 
 |`blobNameFilterValue`
-|The full or partial blob name to be matched.
-It should be specified along with `blobNameFilterRule`.
+|The value to use by `blobNameFilterRule`. The filter is applied to the full blob name (including prefix specified in `blobNamePrefix`), 
+e.g. regular expression applied with `matches` rule must match the full blob name.
 |`.*`
 |===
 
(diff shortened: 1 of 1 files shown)


docs/modules/plugins/pages/plugin-azure-storage-account.adoc
```diff
@@ -239,8 +239,8 @@ It should be specified along with `blobNameFilterValue`.
 |`MATCHES`
 
 |`blobNameFilterValue`
-|The full or partial blob name to be matched.
-It should be specified along with `blobNameFilterRule`.
+|The value to use by `blobNameFilterRule`. The filter is applied to the full blob name (including prefix specified in `blobNamePrefix`),
```
valfirst, line 242:
```suggestion
|The value to use by `blobNameFilterRule`. The filter is applied to the full blob name (including prefix specified in `blobNamePrefix`),
```

uarlouski (approved):

PR #4024, opened 2023-06-07
[plugin-accessiblity] Remove redundant symbol from axe report

uarlouski:


valfirst (approved):

valfirst:
in general it makes sense to fix a typo, and then do refactoring

PR #4025, opened 2023-06-07
[plugin-accessibility] Improve messages for axe a11y testing

uarlouski:


vividus-plugin-accessibility/src/main/java/org/vividus/converter/ParametersToAxeCheckOptionsConverter.java
```diff
@@ -60,8 +60,8 @@ public AxeCheckOptions convertValue(Parameters row, Type type)
         }
 
         AxeCheckOptions checkOptions = new AxeCheckOptions();
-        AxeOptions options = standard != null ? new AxeOptions("tag", List.of(standard.toLowerCase()))
-                : new AxeOptions("rule", violationsToCheck);
+        AxeOptions options = standard != null ? AxeOptions.forStandard(standard.toLowerCase())
```
valfirst, line 63: in this case `toLowerCase` should be moved to factory methods `forStandard `

PR #4027, opened 2023-06-07
[vividus] Add ability to replace deprecated steps automatically

draker94:
(+ replaced steps in stories for debug)


vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -0,0 +1,231 @@
...
+    {
+        Matcher matcher = NESTED_STEPS_TABLE_PATTERN.matcher(parameter);
+        if (matcher.find())
+        {
+            ExamplesTableFactory examplesTableFactory = configuration.examplesTableFactory();
+            ExamplesTable examplesTable = examplesTableFactory.createExamplesTable(parameter);
+
+            String tableContent = examplesTable.getRows().stream().map(p -> p.get(MANDATORY_STEP_COLUMN_NAME))
+                    .collect(Collectors.joining(System.lineSeparator()));
+            List<String> tableNestedSteps = configuration.storyParser().parseStory(tableContent).getScenarios().get(0)
```
valfirst, line 206: I see some similarities to `StepExamplesTableParser`

valfirst (changes requested):
Some TODOs:
- think of integration/system test(s)
- create PR to build system repo
- document Gradle task
- document overall process of deprecating steps (in Developer Guides)
  - how to deprecate regular step
  - how to deprecate composite step
  - describe the ways to find deprecated steps (logging, check gradle task printing steps)
  - describe process of deprecated steps removal

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
draker94: > I see some similarities to `StepExamplesTableParser`

Yep, I took the logic from there partially
But method of the class return `Step` objects and we can't reuse it

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    private void replaceDeprecatedSteps(String testResourcesLocation) throws IOException
+    {
+        IPathFinder pathFinder = BeanFactory.getBean(IPathFinder.class);
+        InjectableStepsFactory stepsFactory = BeanFactory.getBean(InjectableStepsFactory.class);
+
+        BatchConfiguration batchConfiguration = new BatchConfiguration();
+        batchConfiguration.setResourceLocation(Path.of(testResourcesLocation).toUri().toString());
+        batchConfiguration.setResourceIncludePatterns("**/*.story,**/*.steps");
```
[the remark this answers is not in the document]
valfirst: we need to add usage for this class, use Apache Commons CLI and check similar classes for Gradle tasks

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -0,0 +1,231 @@
...
+        batchConfiguration.setResourceExcludePatterns("");
```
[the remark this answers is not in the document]
draker94: No, actually we don't need `ExcludePatterns`

docs/modules/developer-guides/pages/deprecating-steps.adoc
```diff
@@ -0,0 +1,127 @@
...
+Usually `MAJOR` version is incremented on breaking changes releases according to the https://semver.org/spec/v2.0.0.html[Semantic Versioning].
+But if project in initial development phase (has `0.y.z` version) incremented only `MINOR` version.
+
+image::semantic-versions.png[Semantic Versioning]
+
+Also, breaking changes releases usually have the `Removed` section with changes in the https://github.com/vividus-framework/vividus/releases[release history].
+
+image::removed-section.png[Removed section]
+
+In case of update to version with breaking changes necessary to check presence of all removed steps in test project and replace them manually if this cannot be done xref:commons:cli.adoc#_replace_deprecated_steps[automatically].
```
valfirst, line 127:
```suggestion
In case of update to version with breaking changes it is necessary to check presence of all removed steps in test project and replace them manually if this cannot be done xref:commons:cli.adoc#_replace_deprecated_steps[automatically].
```

vividus/src/main/java/org/vividus/steps/DeprecatedSteps.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package org.vividus.steps;
+
+import org.jbehave.core.annotations.Alias;
+import org.jbehave.core.annotations.Given;
+import org.vividus.annotation.Replacement;
+
+public class DeprecatedSteps
```
valfirst, line 23: it should be moved to `vividus-tests` project

valfirst (changes requested):

vividus-plugin-rest-api/src/main/java/org/vividus/steps/api/HttpResponseValidationSteps.java
```diff
@@ -124,7 +128,7 @@ public void doesResponseContainNoBody()
      * @param content        The body content part
      */
     @Deprecated(since = "0.5.0", forRemoval = true)
-    @Replacement(versionToRemoveStep = "0.6.0", replacementFormatPattern = "Then `${response}` is %1$s `%2$s`")
+    @Replacement(versionToRemoveStep = "0.6.0", replacementFormatPattern = "Then `${response}` matches `.*%2$s.*`")
```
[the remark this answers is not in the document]
draker94: > it looks like this step can't be replaced automatically (there is a chance expected response body will contain special symbols or so, and validation on `matches` is not the same as `equals to`)

As far as I understand, we do not have a replacement for this step at all.
Perhaps it shouldn't be under-repaired at the moment at all?

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -0,0 +1,302 @@
...
+                resourceWithReplacedDeprecatedSteps = replaceStepsInSection(resourceWithReplacedDeprecatedSteps,
+                        resourceName, resourceSection.getKey(), resourceSection.getValue(), 0);
+            }
+
+            File file = fullPath.toFile();
+            FileUtils.writeStringToFile(file, resourceWithReplacedDeprecatedSteps, StandardCharsets.UTF_8);
+        }
+        catch (IOException e)
+        {
+            throw new UncheckedIOException(e);
```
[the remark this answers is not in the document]
draker94: > is it really needed?

This method invokes in `forEach` so we must or handle it or replace `forEach` with `for`

vividus-tests/src/main/resources/story/replacement/ReplacedStepsExpected.story
```diff
@@ -0,0 +1,19 @@
+Scenario: Test replacing of deprecated step without parameters
```
uarlouski, line 1: please add metatags, some spacings between steps, maybe examples table, just to make sure it replaces only expected things and has no side effects

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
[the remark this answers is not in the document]
valfirst: It's better to use old-fashioned `for` in this case

valfirst (approved):

.github/workflows/gradle.yml
```diff
@@ -123,7 +123,6 @@ jobs:
       run: |
          ./gradlew validateKnownIssues printSteps countSteps -x testVividusInitialization  -p vividus-tests --project-prop 'vividus.variables.iterationLimit=3' --project-prop 'vividus.variables.target-platform=ios'
 
-
```
valfirst, line 126: noisy change

valfirst (approved):

vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -0,0 +1,303 @@
...
+            });
+            Story story = storyParser.parseStory(resourceContent);
+            Lifecycle lifecycle = story.getLifecycle();
+            steps = story.getScenarios().stream().collect(Collectors.toMap(e -> "Scenario: " + e.getTitle(),
+                    Scenario::getSteps,
+                    (left, right) -> Stream.concat(left.stream(), right.stream()).collect(Collectors.toList())));
+            steps.put("lifecycle (before steps)", lifecycle.getScopes().stream().map(lifecycle::getBeforeSteps)
+                    .flatMap(List::stream).collect(Collectors.toList()));
+            steps.put("lifecycle (after steps)", lifecycle.getScopes().stream().map(lifecycle::getBeforeSteps)
+                    .flatMap(List::stream).collect(Collectors.toList()));
```
uarlouski, line 152: why do we invoke lifecycle::getBeforeSteps on both lines?

uarlouski (approved):

PR #4042, opened 2023-06-12
[plugin-accessibility] Fail if axe check didn't return any results

uarlouski:


ikalinin1 (approved):

vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
```diff
@@ -116,6 +116,15 @@ public void performAccessibilityScan(ExamplesTable checkOptions)
                         AxeReportEntry.class);
 
                 AxeOptions axeOptions = options.getRunOnly();
+                int numberOfResults = reportEntries.stream().map(AxeReportEntry::getResults).map(List::size)
+                        .mapToInt(Integer::intValue).sum();
+                if (numberOfResults == 0)
+                {
+                    softAssert.recordFailedAssertion("Axe Core scan didn't return any results, possibly invalid "
```
valfirst, line 123: I think it should be broken step

vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
```diff
@@ -116,6 +116,15 @@ public void performAccessibilityScan(ExamplesTable checkOptions)
...
+                            + "configuration, please check documentation...");
```
valfirst, line 124: - do not add `...`
- add a hint, like, `provided standard/rules were ..`

valfirst (changes requested):

vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
uarlouski: are you sure? I have doubts because we may provide more than one configuration for the step so we may not get any result if invalid config is at the beginning of the configuration list

valfirst: invalid config is an user error, it must be never committed

valfirst (approved):

PR #4062, opened 2023-06-20
[vividus] Mark deprecated composite steps in printing steps task

### LINKED ISSUE #4048: Improve `printSteps` to print deprecation message for composite steps
Improve `printSteps` to print deprecation message for composite steps

_Originally posted by @valfirst in https://github.com/vividus-framework/vividus/pull/4027#discussion_r1230564110_

draker94:
Closes #4048

### DIFF
--- a/docs/modules/developer-guides/pages/deprecating-steps.adoc
+++ b/docs/modules/developer-guides/pages/deprecating-steps.adoc
@@ -41,13 +41,13 @@ org.vividus.replacement.DeprecatedCodeStepsReporter - The step: "Given the step 
 == Deprecate composite steps
 
 For deprecation of composite steps use the comment with the special pattern:
-`!--\s+Deprecated:\s+(.*),\s+(.*)$` where the first group is version to remove, the second one is formatter pattern for the actual step.
+`!--\s+DEPRECATED:\s+(.*),\s+(.*)$` where the first group is version to remove, the second one is formatter pattern for the actual step.
 
 .Deprecated composite step in `composite.steps` file
 [source,gherkin]
 ----
 Composite: Given the step with parameters '$parameter1' and '$parameter2'
-!-- Deprecated: 0.6.0, Given the step with parameter `%2$s` and `%1$s`
+!-- DEPRECATED: 0.6.0, Given the step with parameter `%2$s` and `%1$s`
 Given the step with parameters `<parameter2>` and `<parameter1>`
 ----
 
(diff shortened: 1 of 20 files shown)


vividus/src/main/java/org/vividus/runner/DeprecatedStepsReplacer.java
```diff
@@ -75,7 +75,7 @@ public final class DeprecatedStepsReplacer
     private static final String END_OF_STEP_LOOKAHEAD = "(?=\\h*(\\v|$))";
     private static final Pattern NESTED_STEPS_TABLE_PATTERN = Pattern.compile("(\\v|^).\\h*step\\h*.\\v");
     private static final Pattern COMPOSITE_DEPRECATION_NOTIFICATION
-            = Pattern.compile("!--\\s+WARNING: The step .* is deprecated and will be removed in .*");
+            = Pattern.compile("!--\\s+Deprecated: The step .* is deprecated and will be removed in .*");
```
uarlouski, line 78: can we do this in upper case or case insensitive? it's just more visible this way

draker94: > can we do this in upper case or case insensitive? it's just more visible this way

Added `CASE_INSENSITIVE` flags.

draker94: > > can we do this in upper case or case insensitive? it's just more visible this way
> 
> Added `CASE_INSENSITIVE` flags.

But actually it doesn't make any sense to do  deprecation pattern case insensitive. 
I think in this place better to leave only the uppercase

docs/modules/developer-guides/pages/deprecating-steps.adoc
```diff
@@ -59,13 +59,13 @@ org.vividus.log.LoggingStoryReporter - The step: "Given the step with parameters
 
 [NOTE]
 ====
-If composite step cannot be deprecated using formatter pattern it still must have the comment with the certain pattern for user notification in step replacer: `!--\s+WARNING: The step .* is deprecated and will be removed in .*`.
+If composite step cannot be deprecated using formatter pattern it still must have the comment with the certain pattern for user notification in step replacer: `!--\s+DEPRECATED: The step .* is deprecated and will be removed in .*`.
```
valfirst, line 62: I guess you should update all composite steps to follow new pattern

vividus/src/main/java/org/vividus/runner/StepsPrinter.java
```diff
@@ -49,6 +51,8 @@ public class StepsPrinter
     private static final String DEPRECATED = "DEPRECATED";
     private static final String COMPOSITE = "COMPOSITE IN STEPS FILE";
     private static final String EMPTY = "";
+    private static final Pattern DEPRECATION_COMMENT_PATTERN = Pattern.compile("^!--\\h+Deprecated:\\h+.*",
+            Pattern.CASE_INSENSITIVE);
```
valfirst, line 55: I don't think it should be case insensitive, markers format should be strict: only upper case

valfirst (changes requested):

valfirst (approved):

PR #4127, opened 2023-07-07
[plugin-mongodb] Fix serialization of ObjectId

valfirst:


vividus-plugin-mongodb/src/main/java/org/vividus/mongodb/databind/MongoDbJacksonMapperModule.java
```diff
@@ -0,0 +1,43 @@
...
+public class MongoDbJacksonMapperModule extends SimpleModule
+{
+    private static final long serialVersionUID = -8454324208644523711L;
+
+    public MongoDbJacksonMapperModule()
+    {
+        addSerializer(ObjectId.class, new JsonSerializer<>()
+        {
+            @Override
+            public void serialize(ObjectId value, JsonGenerator gen, SerializerProvider serializers) throws IOException
```
ikalinin1, line 37:
```suggestion
            public void serialize(ObjectId value, JsonGenerator generator, SerializerProvider serializers) throws IOException
```

ikalinin1 (approved):

vividus-plugin-mongodb/src/main/java/org/vividus/mongodb/databind/MongoDbJacksonMapperModule.java
valfirst: https://github.com/FasterXML/jackson-databind/blob/cbd5c35a028953fab3962960a75bf8df86fe001e/src/main/java/com/fasterxml/jackson/databind/JsonSerializer.java#L122

uarlouski (approved):

ikalinin1:
I think we need to cover this with integration test, if possible

PR #4130, opened 2023-07-11
Use `ServiceLoader` mechanism to register Jackson modules

valfirst:


vividus-util/src/test/java/org/vividus/util/json/JsonUtilsTests.java
```diff
@@ -160,6 +161,14 @@ void shouldReadJson()
             ).toPrettyString());
     }
 
+    @Test
+    void shouldCreatePattern()
```
ikalinin1, line 165: Is the logic covered with this test obsolete now? 

```java
    @Test
    void shouldThrowExceptionInCaseOfNullPattern()
    {
        JsonParser jsonParser = mock(JsonParser.class);
        IllegalArgumentException exception = assertThrows(IllegalArgumentException.class,
            () -> new PatternDeserializer().deserialize(jsonParser, null).toString());
        assertEquals("Pattern could not be empty", exception.getMessage());
    }
```

valfirst: I've checked this case during implementation, it can't happen anymore

uarlouski (approved):

PR #4141, opened 2023-07-14
[plugin-json] Fix wording of deprecated JSON nested step

draker94:


vividus-plugin-json/src/main/resources/steps/defaults/deprecated.steps
```diff
@@ -19,7 +19,7 @@ Priority: 6
 !-- DEPRECATED: 0.6.0, When I find %1$s `%2$s` JSON elements from `${json-context}` by `%3$s` and for each element do%4$s
 When I find <comparisonRule> `<elementsNumber>` JSON elements from `${json-context}` by `<jsonPath>` and for each element do<stepsToExecute>
 
-Composite: When I find $comparisonRule '$elementsNumber' JSON elements by '$jsonPath' and for each element doa$stepsToExecute
+Composite: When I find $comparisonRule '$elementsNumber' JSON elements by '$jsonPath' and for each element do$stepsToExecute
```
valfirst, line 22: this is a kind of expected and known typo: it was in the original step, we deprecated it, but we couldn't change its syntax, because there would be breaking changes

this step will be removed in 0.6.0

PR #4150, opened 2023-07-18
[engine] Add ability to disable resource loader configuration

draker94:


vividus/src/main/resources/properties/defaults/default.properties
```diff
@@ -6,6 +6,7 @@ bdd.meta-filters=true
 # The property default value currently specified in org.vividus.batch.BatchStorage in order to support conflicting properties check.
 # story.execution-timeout=PT3H
 bdd.cache-examples-table=false
+bdd.enable-resource-loader-configuration=false
```
valfirst, line 9: we should keep it turned on in `0.5.X`

vividus-engine/src/main/resources/org/vividus/engine/spring.xml
```diff
@@ -38,7 +38,9 @@
 
     <bean class="org.vividus.parser.StepExamplesTableParser" />
 
-    <bean id="testResourceLoader" class="org.vividus.resource.TestResourceLoader" />
+    <bean id="testResourceLoader" class="org.vividus.resource.TestResourceLoader">
+        <constructor-arg name="enableResourceLoaderConfiguration" value="${bdd.enable-resource-loader-configuration}" />
```
valfirst, line 42:
```suggestion
        <constructor-arg name="enableResourceLoaderConfiguration" value="${engine.dynamic-resource-search-enabled}" />
```

valfirst (changes requested):

vividus-engine/src/main/java/org/vividus/resource/TestResourceLoader.java
```diff
@@ -40,12 +40,18 @@ public class TestResourceLoader implements ITestResourceLoader
...
     private final ResourcePatternResolver resourcePatternResolver;
 
-    public TestResourceLoader(IPropertyParser propertyParser, ResourcePatternResolver resourcePatternResolver)
+    private final boolean enableResourceLoaderConfiguration;
+
+    public TestResourceLoader(IPropertyParser propertyParser, ResourcePatternResolver resourcePatternResolver,
+            boolean enableResourceLoaderConfiguration)
     {
-        resourceLoadParameters = propertyParser.getPropertyValuesByPrefix(VARIABLES_PROPERTY_PREFIX).values().stream()
+        resourceLoadParameters = enableResourceLoaderConfiguration
```
valfirst, line 48: also it makes to log some warning, that this feature is deprecated and will be removed in 0.7.0

valfirst (changes requested):

valfirst (approved):

vividus-engine/src/main/resources/org/vividus/engine/spring.xml
```diff
@@ -38,7 +38,9 @@
 
     <bean class="org.vividus.parser.StepExamplesTableParser" />
 
-    <bean id="testResourceLoader" class="org.vividus.resource.TestResourceLoader" />
+    <bean id="testResourceLoader" class="org.vividus.resource.TestResourceLoader">
+        <constructor-arg name="dynamicResourceSearchEnabled" value="${engine.dynamic-resource-search-enabled}" />
```
uarlouski, line 42: it works only because there is only on boolean in the constructor, the args names are not preserved by default, so I would use index for safety

uarlouski (approved):

PR #4208, opened 2023-08-08
[plugin-web-app] Remove the majority of steps using explicit XPaths as locators

draker94:


vividus-tests/src/main/resources/story/integration/GenericSetVariableSteps.story
```diff
@@ -54,5 +54,5 @@ Scenario: Verify save number of elements steps
 Given I am on page with URL `${vividus-test-site-url}/links.html`
 When I save number of elements located `tagName(a)` to SCENARIO variable `numberOfLinks`
 Then `${numberOfLinks}` is equal to `2`
-When I set the number of elements found by xpath '//a' to the 'SCENARIO' variable 'numberOfLinks'
+When I save number of elements located `By.xpath(//a)` to SCENARIO variable `numberOfLinks`
 Then `${numberOfLinks}` is equal to `2`
```
valfirst, line 58: I think this can be removed as we check this step above

valfirst (approved):

PR #4213, opened 2023-08-09
[vividus] Remove deprecated step validating elements sorting order

draker94:


valfirst (changes requested):
It make sense to update docs as well

valfirst (approved):

PR #4217, opened 2023-08-10
[plugin-web-app] Do not require Windows for MS Edge browser

valfirst:


valfirst:
SauceLabs requires explicit OS configuration

PR #4227, opened 2023-08-14
[plugin-applitools] Remove config dependency on plugin rest api

uarlouski:


vividus-plugin-applitools/src/main/resources/vividus-plugin/spring.xml
```diff
@@ -37,7 +37,7 @@
         factory-bean="httpClientFactory"
         factory-method="buildHttpClient">
         <constructor-arg>
-            <bean parent="restApiHttpClientConfig">
+            <bean parent="httpClientConfig">
```
valfirst, line 40: In this case I believe the configuration of interceptors and handlers becomes redundant

PR #4241, opened 2023-08-17
Fix deprecated step replacer for steps with percent char

vkepin:


vividus-engine/src/main/java/org/vividus/replacement/DeprecatedStepNotificationFactory.java
```diff
@@ -49,9 +50,10 @@ protected String createDeprecatedStepNotification(String versionToRemove, String
         String[] stepParams = getParametersFromStep(stepMatcher);
 
         String notification;
+        String normalizedFormatPattern = replacementFormatPattern.replaceAll(PERCENT_ESCAPE_PATTERN, "%%");
```
valfirst, line 53: actually we need to fix replacement patterns with `%` instead

PR #4357, opened 2023-09-25
[plugin-accessibility] Align target field deserialization with axe core rules

uarlouski:
Accessibility check is broken on listerine.com and tyenol.com sites because target field deserialization doesn't cover all output forms of target field described here: https://www.deque.com/axe/core-documentation/api-documentation/
```
target - Array of either strings or Arrays of strings. If the item in the array is a string, then it is a CSS selector. 
If there are multiple items in the array each item corresponds to one level of iframe or frame. If there is one iframe or frame, there should be two entries in target. 
If there are three iframe levels, there should be four entries in target. 
If the item in the Array is an Array of strings, then it points to an element in a shadow DOM and each item (except the n-1th) in this array is a selector to a DOM element with a shadow DOM. 
The last element in the array points to the final shadow DOM node.
```

### DIFF
--- a/vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/deserializer/TargetDeserializer.java
+++ b/vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/deserializer/TargetDeserializer.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/deserializer/TargetDeserializer.java
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.vividus.accessibility.model.axe.Target;
+
+public class TargetDeserializer extends JsonDeserializer<Target>
+{
+    @Override
+    public Target deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JacksonException
```
valfirst, line 34:
```suggestion
    public Target deserialize(JsonParser p, DeserializationContext ctxt) throws IOException
```

valfirst (changes requested):

valfirst (approved):

PR #4368, opened 2023-09-27
[util] Prevent iteration waiting in duration waiter if condition reached

draker94:


valfirst (approved):

vividus-util/src/main/java/org/vividus/util/wait/DurationBasedWaiter.java
```diff
@@ -50,14 +50,18 @@ public <T, E extends Exception> T wait(FailableSupplier<T, E> valueProvider, Pre
             long iterationStartTime = System.currentTimeMillis();
             value = valueProvider.get();
             long iterationEndTime = System.currentTimeMillis();
+            if (stopCondition.test(value))
+            {
+                break;
+            }
```
valfirst, line 56:
```suggestion
            if (stopCondition.test(value))
            {
                break;
            }
            long iterationEndTime = System.currentTimeMillis();
```

draker94: This is how it was at the beginning
And we again get **this** error

image

valfirst: I guess this test is needed to be tuned

valfirst: this is expected, since we're changing the underneath logic

uarlouski (approved):

valfirst (approved):

PR #4386, opened 2023-09-30
[util] Percent-encode all restricted characters in URL path

draker94:


vividus-util/src/test/java/org/vividus/util/UriUtilsTests.java
```diff
@@ -148,18 +148,18 @@ void testRemoveUnexistentFragment()
         // CHECKSTYLE:OFF
         "http://somehost:8080/path,                                   http://somehost:8080/path",
         "http://somehost:8080/path?name=goodvalue&a=b#fragment,       http://somehost:8080/path?name=goodvalue&a=b#fragment",
-        "http://somehost:8080/pa | th?name=bad|value&a=b#fra| gme nt, http://somehost:8080/pa%20%7C%20th?name=bad%7Cvalue&a=b#fra%7C%20gme%20nt",
+        "http://somehost:8080/pa | th?name=bad|value&a=b#fra| gme nt, http://somehost:8080/pa+%7C+th?name=bad%7Cvalue&a=b#fra%7C%20gme%20nt",
```
valfirst, line 151: I'm not sure it's correct to replace space with `+` we need to conform this change with RFC

vividus-util/src/test/java/org/vividus/util/UriUtilsTests.java
```diff
@@ -148,18 +148,18 @@ void testRemoveUnexistentFragment()
...
         "tel:1234567#0987,                                            tel:1234567#0987",
         "https://ad.doubleclick.net/ddm/activity/src=5337729;type=brand0;cat=brand0;u9=[Cachebuster];u10=[SPIKA Locale];u11=[SPIKA Brand];u12=[Page Path];u13=[SPIKA Language];u14=[Cocktail ID];u15=[Page Type];dc_lat=;dc_rdid=;tag_for_child_directed_treatment=;tfua=;npa=;ord=1?," +
-            "https://ad.doubleclick.net/ddm/activity/src=5337729;type=brand0;cat=brand0;u9=%5BCachebuster%5D;u10=%5BSPIKA%20Locale%5D;u11=%5BSPIKA%20Brand%5D;u12=%5BPage%20Path%5D;u13=%5BSPIKA%20Language%5D;u14=%5BCocktail%20ID%5D;u15=%5BPage%20Type%5D;dc_lat=;dc_rdid=;tag_for_child_directed_treatment=;tfua=;npa=;ord=1?",
+            "https://ad.doubleclick.net/ddm/activity/src%3D5337729%3Btype%3Dbrand0%3Bcat%3Dbrand0%3Bu9%3D%5BCachebuster%5D%3Bu10%3D%5BSPIKA+Locale%5D%3Bu11%3D%5BSPIKA+Brand%5D%3Bu12%3D%5BPage+Path%5D%3Bu13%3D%5BSPIKA+Language%5D%3Bu14%3D%5BCocktail+ID%5D%3Bu15%3D%5BPage+Type%5D%3Bdc_lat%3D%3Bdc_rdid%3D%3Btag_for_child_directed_treatment%3D%3Btfua%3D%3Bnpa%3D%3Bord%3D1?",
```
valfirst, line 156: is `=` required to be percent-encoded?

valfirst (changes requested):
In general, I think the changes should be less aggressive - we need to percent-encode only non-ASCII characters

vividus-util/src/main/java/org/vividus/util/UriUtils.java
```diff
@@ -296,8 +317,11 @@ public static URI buildNewUrl(URI url, String relativeUrl)
             }
 
             String path = StringUtils.repeat(SLASH, indexOfFirstNonSlashChar - 1) + parsedRelativeUrl.getPath();
```
valfirst, line 319: can we just encode this path? (instead of creating 2 instances of `URI`)

valfirst (changes requested):

valfirst (changes requested):
Please, make sure to cover all the changes and branches with unit tests

draker94:
> Please, make sure to cover all the changes and branches with unit tests

The exception below can't be covered because we never get `URISyntaxException` in this method (the method already get URI as param and creating URI from path can get to us only `IllegalArgumentException`). But we intercept it to avoid the need to handle this exception in each method call.
https://github.com/vividus-framework/vividus/pull/4386/files#diff-971440b884468d146e7513151e0742bbf01f2b2d0f52fdf3363792cdccfbae9cR329

valfirst (approved):

uarlouski (approved):

PR #4432, opened 2023-10-12
[plugin-lighthouse] Optimize unit tests execution

uarlouski:


vividus-plugin-lighthouse/src/test/java/org/vividus/lighthouse/LighthouseStepsTests.java
```diff
@@ -535,6 +541,16 @@ private LighthouseResultV5 createLighthouseResult(double performance, double seo
         return results;
     }
 
+    private static void mockSleeper(FailableRunnable<Exception> runnable, int times) throws Exception
```
valfirst, line 544:
```suggestion
    private static void runWithMockedSleeper(FailableRunnable<Exception> runnable, int times) throws Exception
```

valfirst (approved):

PR #4534, opened 2023-11-06
[allure-adaptor] Add ability to hide Parameters section in report

### LINKED ISSUE #1047: Add a possibility to enable/disable 'parameters' section display in Allure
In Allure there is a section Suites > Overview > Parameters where all input data generated for the run and stored in .table file 
is displayed.

If .table file contains a lot of records, parameters section tends to be super long and its hard to navigate to the actual Execution body.

The task is to add a possibility to disable 'parameters' section or 'collapse by default' so that navigation is more user friendly.

valfirst:
Closes #1047

### DIFF
--- a/vividus/src/main/resources/properties/defaults/default.properties
+++ b/vividus/src/main/resources/properties/defaults/default.properties
@@ -86,4 +86,5 @@ template-processor.resolve-bdd-variables=false
 statistics.folder=${output.directory}/statistics
 statistics.print-failures=false
 
+report.show-parameters-section=true
 report.text-length-diff-threshold=100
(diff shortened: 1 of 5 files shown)


vividus/src/main/resources/properties/defaults/default.properties
```diff
@@ -86,4 +86,5 @@ template-processor.resolve-bdd-variables=false
 statistics.folder=${output.directory}/statistics
 statistics.print-failures=false
 
+report.show-parameter-section=true
```
uarlouski, line 89: parameters?

uarlouski (approved):

PR #4537, opened 2023-11-06
[plugin-rest-api] Add extended logging for HTTP requests and responses

### LINKED ISSUE #4437: Request and response details in the execution log
During REST API test development, it would be great to see the request and response details (eg headers, actual response body etc) to be able to quickly see whats could be wrong. 

This information is available in the Allure report, but when the tests are running on the pipeline, logs easier to check than downloading the report then opening it. 

There could be a switch to enable-disable this "debug" feature to make sure the logs are not cluttered with this extra info if someone thinks its not useful (what I dont think anyone woul feel :D )

draker94:
Closes #4437

### DIFF
--- a/vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
+++ b/vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2019-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.vividus.http;
+
+import static org.apache.hc.core5.http.ContentType.APPLICATION_JSON;
+import static org.apache.hc.core5.http.ContentType.APPLICATION_XML;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.hc.core5.http.EntityDetails;
+import org.apache.hc.core5.http.Header;
+import org.apache.hc.core5.http.HttpEntity;
+import org.apache.hc.core5.http.HttpEntityContainer;
+import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpRequestInterceptor;
+import org.apache.hc.core5.http.ProtocolVersion;
+import org.apache.hc.core5.http.io.entity.StringEntity;
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.spi.LoggingEventBuilder;
+import org.vividus.http.client.HttpResponse;
+import org.vividus.http.handler.HttpResponseHandler;
+
+public class ExtendedHttpLoggingInterceptor implements HttpRequestInterceptor, HttpResponseHandler
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpRequestExecutor.class);
+
+    private static final String NEW_LINE = System.lineSeparator();
+    private static final String HEADERS_FORMAT = String.format("%nHeaders:%n{}");
+    private static final String BODY_FORMAT = String.format("%nBody:%n{}");
+
+    private static final Set<String> LOGGED_CONTENT_TYPES = Set.of(APPLICATION_JSON.getMimeType(),
+            APPLICATION_XML.getMimeType());
+
+    private final boolean extendedLogging;
+
+    public ExtendedHttpLoggingInterceptor(boolean extendedLogging)
+    {
+        this.extendedLogging = extendedLogging;
+    }
+
+    @Override
+    public void process(HttpRequest request, EntityDetails entityDetails, HttpContext context) throws IOException
+    {
+        String argBrackets = " {}";
+        LoggingEventBuilder loggingEventBuilder = LOGGER.atInfo();
+        StringBuilder loggerFormat = new StringBuilder("Request:");
+
+        ProtocolVersion protocolVersion = request.getVersion();
+        if (protocolVersion != null)
+        {
+            loggingEventBuilder = loggingEventBuilder.addArgument(protocolVersion);
+            loggerFormat.append(argBrackets);
+        }
+        loggingEventBuilder = loggingEventBuilder.addArgument(request);
+        loggerFormat.append(argBrackets);
+
+        if (extendedLogging)
+        {
+            loggingEventBuilder = loggingEventBuilder.addArgument(
+                    () -> Stream.of(request.getHeaders()).map(Object::toString).collect(Collectors.joining(NEW_LINE)));
+            loggerFormat.append(HEADERS_FORMAT);
+
+            if (request instanceof HttpEntityContainer httpEntityContainer)
+            {
+                HttpEntity entity = httpEntityContainer.getEntity();
+                if (entity instanceof StringEntity stringEntity)
+                {
+                    loggingEventBuilder = loggingEventBuilder.addArgument(() ->
+                    {
+                        try
+                        {
+                            return new String(stringEntity.getContent().readAllBytes(), StandardCharsets.UTF_8);
+                        }
+                        catch (IOException e)
+                        {
+                            return "Unable to get body content: " + e.getMessage();
+                        }
+                    });
+                    loggerFormat.append(BODY_FORMAT);
+                }
+                else if (entity != null)
+                {
+                    int bodySizeInBytes = entity.getContent().available();
+                    loggingEventBuilder = loggingEventBuilder.addArgument(bodySizeInBytes);
+                    loggerFormat.append(NEW_LINE).append("Body: {} bytes of binary data");
+                }
+            }
+        }
+        loggingEventBuilder.log(loggerFormat.toString());
+    }
+
+    @Override
+    public void handle(HttpResponse httpResponse)
+    {
+        LoggingEventBuilder loggingEventBuilder = LOGGER.atInfo().addArgument(httpResponse.getStatusCode())
+                .addArgument(httpResponse.getFrom());
+        StringBuilder loggerFormat = new StringBuilder("Response: status code {}, {}");
+
+        if (extendedLogging)
+        {
+            Header[] headers = httpResponse.getResponseHeaders();
+            loggingEventBuilder = loggingEventBuilder
+                    .addArgument(() -> Stream.of(headers).map(Object::toString).collect(Collectors.joining(NEW_LINE)));
+            loggerFormat.append(HEADERS_FORMAT);
+
+            if (httpResponse.getResponseBody() != null)
+            {
+                String mimeType = MimeTypeUtils.getMimeTypeFromHeadersWithDefault(headers);
+                if (mimeType.startsWith("text/") || LOGGED_CONTENT_TYPES.contains(mimeType))
+                {
+                    loggingEventBuilder = loggingEventBuilder.addArgument(httpResponse::getResponseBodyAsString);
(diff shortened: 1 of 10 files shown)


vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public void process(HttpRequest request, EntityDetails entityDetails, HttpContext context) throws IOException
+    {
+        String argBrackets = " {}";
+        LoggingEventBuilder loggingEventBuilder = logger.atInfo();
+        StringBuilder loggerFormat = new StringBuilder("Request:");
+
+        ProtocolVersion protocolVersion = request.getVersion();
+        if (protocolVersion != null)
```
valfirst, line 79: is it possible to have no version?

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+            if (LOGGED_CONTENT_TYPES.contains(mimeType) && httpResponse.getResponseBody() != null)
+            {
+                loggingEventBuilder = loggingEventBuilder.addArgument(httpResponse.getResponseBodyAsString());
+                loggerFormat.append(BODY_FORMAT);
+            }
+        }
+        loggingEventBuilder.log(loggerFormat.toString());
+    }
+
+    private Optional<String> getMimeType(Header... headers)
```
valfirst, line 132: the logic is a partial copy of `PublishingAttachmentInterceptor`, please refactor to avoid duplications

valfirst (changes requested):

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
draker94: > is it possible to have no version?

It seems like it usually null
image

valfirst: probably the request is intercepted too early

draker94: > probably the request is intercepted too early

In debug the protocol version is set only for response, but not for request
Looks like the protocol version needs to be set manually
During req execution HTTP client using HTTP 1.1 as default protocol for request and context
https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/protocol/RequestExpectContinue.java#L69
https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/BasicHttpContext.java#L95

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+import org.slf4j.LoggerFactory;
+import org.slf4j.spi.LoggingEventBuilder;
+import org.vividus.http.client.HttpResponse;
+import org.vividus.http.handler.HttpResponseHandler;
+
+public class ExtendedHttpLoggingInterceptor implements HttpRequestInterceptor, HttpResponseHandler
+{
+    private static final String NEW_LINE = "\n";
+    private static final String HEADERS_FORMAT = "\nHeaders:\n{}";
+    private static final String BODY_FORMAT = "\nBody:\n{}";
```
uarlouski, line 49: in general isn't is better to log HTTP request/response in a format described in spec? like
image

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+        {
+            Header[] headers = request.getHeaders();
+            String headersAsString = Stream.of(headers).map(Object::toString).collect(Collectors.joining(NEW_LINE));
+            loggingEventBuilder = loggingEventBuilder.addArgument(headersAsString);
+            loggerFormat.append(HEADERS_FORMAT);
+
+            if (request instanceof HttpEntityContainer httpEntityContainer)
+            {
+                HttpEntity entity = httpEntityContainer.getEntity();
+                if (entity instanceof StringEntity stringEntity)
```
uarlouski, line 90: if the body is not a string should we add some record like <...binary content> or smth to log? so we can distinguish if the body is empty or not?

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.hc.core5.http.protocol.HttpContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.spi.LoggingEventBuilder;
+import org.vividus.http.client.HttpResponse;
+import org.vividus.http.handler.HttpResponseHandler;
+
+public class ExtendedHttpLoggingInterceptor implements HttpRequestInterceptor, HttpResponseHandler
+{
+    private static final String NEW_LINE = "\n";
```
uarlouski, line 47: System.lineSeparator()?

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+        {
+            loggingEventBuilder = loggingEventBuilder.addArgument(protocolVersion);
+            loggerFormat.append(argBrackets);
+        }
+        loggingEventBuilder = loggingEventBuilder.addArgument(request);
+        loggerFormat.append(argBrackets);
+
+        if (extendedLogging)
+        {
+            Header[] headers = request.getHeaders();
```
uarlouski, line 82: can we inline it here and below?

draker94: > can we inline it here and below?

Here can below not (using twice)

valfirst (changes requested):

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    String textBody = new String(stringEntity.getContent().readAllBytes(), StandardCharsets.UTF_8);
+                    loggingEventBuilder = loggingEventBuilder.addArgument(textBody);
+                    loggerFormat.append(BODY_FORMAT);
+                }
+                else if (entity != null)
+                {
+                    int bodySizeInBytes = entity.getContent().available();
+                    loggingEventBuilder = loggingEventBuilder.addArgument(bodySizeInBytes);
+                    loggerFormat.append(NEW_LINE).append("Body: not text data. Size {} bytes.");
```
valfirst, line 99:
```suggestion
                    loggerFormat.append(NEW_LINE).append("Body: {} bytes of binary data");
```

vividus-plugin-rest-api/src/main/java/org/vividus/http/MimeTypeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static Optional<String> getMimeTypeFromHeaders(Header... headers)
+    {
+        return Stream.of(headers)
+                .filter(h -> HttpHeaders.CONTENT_TYPE.equalsIgnoreCase(h.getName())
+                        && StringUtils.isNotBlank(h.getValue()))
+                .findFirst()
+                .map(MessageSupport::parse)
+                .map(elements -> elements[0])
+                .map(HeaderElement::getName);
```
valfirst, line 42:
```suggestion
                .map(HeaderElement::getName)
                .orElseGet(ContentType.DEFAULT_TEXT::getMimeType);
```
and then add javadocs describing the current behavior

vividus-plugin-rest-api/src/main/resources/vividus-plugin/spring.xml
```diff
@@ -80,6 +82,11 @@
         <constructor-arg ref="softAssert" />
     </bean>
 
+    <bean id="extendedHttpLoggingInterceptor" class="org.vividus.http.ExtendedHttpLoggingInterceptor">
+        <constructor-arg value="org.vividus.http.HttpRequestExecutor" type="java.lang.Class" />
```
valfirst, line 86: is it really needed to specify type?

valfirst (changes requested):

vividus-plugin-rest-api/src/main/java/org/vividus/http/MimeTypeUtils.java
draker94: > and then add javadocs describing the current behavior

We have one place where we get type with additional logic
https://github.com/vividus-framework/vividus/pull/4537/files#diff-dd131ecfa6f29400ddeb501304099af8e36c2942b87e6b5f47490af5ee1bc919L68
So I decided to create separate utils method

valfirst (approved):

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+                HttpEntity entity = httpEntityContainer.getEntity();
+                if (entity instanceof StringEntity stringEntity)
+                {
+                    String textBody = new String(stringEntity.getContent().readAllBytes(), StandardCharsets.UTF_8);
+                    loggingEventBuilder = loggingEventBuilder.addArgument(textBody);
+                    loggerFormat.append(BODY_FORMAT);
+                }
+                else if (entity != null)
+                {
+                    int bodySizeInBytes = entity.getContent().available();
```
uarlouski, line 95: getContentLength() ?

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+            loggingEventBuilder = loggingEventBuilder.addArgument(headersAsString);
+            loggerFormat.append(HEADERS_FORMAT);
+
+            if (request instanceof HttpEntityContainer httpEntityContainer)
+            {
+                HttpEntity entity = httpEntityContainer.getEntity();
+                if (entity instanceof StringEntity stringEntity)
+                {
+                    String textBody = new String(stringEntity.getContent().readAllBytes(), StandardCharsets.UTF_8);
+                    loggingEventBuilder = loggingEventBuilder.addArgument(textBody);
```
uarlouski, line 90: shouldn't it be supplier?

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+                    loggerFormat.append(BODY_FORMAT);
+                }
+                else if (entity != null)
+                {
+                    int bodySizeInBytes = entity.getContent().available();
+                    loggingEventBuilder = loggingEventBuilder.addArgument(bodySizeInBytes);
```
uarlouski, line 96: shouldn't it be supplier?

uarlouski: please check other places

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
uarlouski: please check other places

vividus-plugin-rest-api/src/main/java/org/vividus/http/ExtendedHttpLoggingInterceptor.java
draker94: > getContentLength

Entities can be created without explicitly specifying length:
https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/io/entity/BasicHttpEntity.java#L67-L69
So I think it's safer to take the length right from the content

uarlouski (approved):

PR #4598, opened 2023-11-24
[plugin-web-app-to-rest-api] Add ability to set headers for headless …

uarlouski:
…crawler requests

### DIFF
--- a/docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
+++ b/docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
@@ -182,6 +182,16 @@ a|`true`
 |`true`
 |Whether to crawl https pages.
 
+|`transformer.from-headless-crawling.http.headers.<header name>=<header value>`
+|
+|
+a|Set of headers to set for every crawling request being sent.
+
+[source,properties]
+----
+transformer.from-headless-crawling.http.headers.x-vercel-protection-bypass=1fac2b25014d35e5103b
+----
+
 4+^.^|_Proxy_
 
 |`transformer.from-headless-crawling.proxy-host`
--- a/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
+++ b/vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
@@ -19,11 +19,15 @@ package org.vividus.crawler;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.util.Collection;
 import java.util.Map;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.hc.core5.http.Header;
 import org.vividus.util.UriUtils;
 import org.vividus.util.UriUtils.UserInfo;
 import org.vividus.util.property.IPropertyMapper;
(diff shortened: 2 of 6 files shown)


docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
```diff
@@ -182,6 +182,16 @@ a|`true`
 |`true`
 |Whether to crawl https pages.
 
+|`transformer.from-headless-crawling.default-headers.<header name>=<header value>`
```
valfirst, line 185:
```suggestion
|`transformer.from-headless-crawling.http.header.<header name>=<header value>`
```

valfirst (changes requested):
Something is missing: no property mapping to any bean field/arg

docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
valfirst: but we override default class, and thus can keep consistent mapping in properties

uarlouski: as to consistency, 90% properties mentioned here https://docs.vividus.dev/vividus/latest/plugins/plugin-web-app-to-rest-api.html#_from_headless_crawling are about http, but we didn't prefix them with http

valfirst: https://github.com/vividus-framework/vividus/blob/4618e8aadc833c49b7e2f3d70b3d7ff8eb2519ca/vividus-plugin-rest-api/src/main/resources/vividus-plugin/spring.xml#L58

https://github.com/vividus-framework/vividus/blob/4618e8aadc833c49b7e2f3d70b3d7ff8eb2519ca/vividus-plugin-web-app-to-rest-api/src/main/resources/vividus-plugin/spring.xml#L59

valfirst, line 185: the doc should be updated now

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
```diff
@@ -97,7 +101,28 @@ private CrawlConfig createCrawlConfig(URI mainApplicationPage) throws IOExceptio
...
+
+        public void setHttp(Http http)
+        {
+            this.http = http;
+        }
+
+        @SuppressWarnings({ "EmptyPublicCtorInClass", "RedundantModifier" })
+        private final class Http
+        {
+            public Http()
```
valfirst, line 118: why is it needed?

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
```diff
@@ -97,7 +101,28 @@ private CrawlConfig createCrawlConfig(URI mainApplicationPage) throws IOExceptio
...
     {
+        private Http http;
+
+        public void setHttp(Http http)
+        {
+            this.http = http;
+        }
+
+        @SuppressWarnings({ "EmptyPublicCtorInClass", "RedundantModifier" })
+        private final class Http
```
valfirst, line 116: avoid mixing visibility modifiers, either make everything `private` or `public`

valfirst: or maybe package-private?

valfirst (changes requested):

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
uarlouski: required by jackson: non-static inner classes like this can only by instantiated using default, no-argument constructor

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
uarlouski: made everything private

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/CrawlControllerFactory.java
valfirst: is package-private an option here?

uarlouski: no, it has to be public

valfirst (approved):

PR #4622, opened 2023-12-01
[plugin-web-app-to-rest-api] Add HTML document table transformer

uarlouski:


vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/crawler/transformer/HtmlDocumentTableTransformer.java
```diff
@@ -0,0 +1,109 @@
...
+        {
+            getter = Element::text;
+            xpathSelector = substringBeforeLast(xpathSelector, SLASH);
+        }
+        else if (lastSegment.startsWith("@"))
+        {
+            String attribute = lastSegment.substring(1);
+            getter = e -> e.attr(attribute);
+            xpathSelector = substringBeforeLast(xpathSelector, SLASH);
+        }
```
valfirst, line 71:
```suggestion
        Function<Element, String> getter;
        if ("text()".equals(lastSegment))
        {
            getter = Element::text;
            xpathSelector = substringBeforeLast(xpathSelector, SLASH);
        }
        else if (lastSegment.startsWith("@"))
        {
            String attribute = lastSegment.substring(1);
            getter = e -> e.attr(attribute);
            xpathSelector = substringBeforeLast(xpathSelector, SLASH);
        }
        else
        {
            getter = Element::outerHtml;
        }
```

valfirst (changes requested):

PR #4638, opened 2023-12-05
[vividus-util] Use getRawAuthority() in buildNewUrl method to get the original encoded authority value

TykovkaV:


vividus-util/src/main/java/org/vividus/util/UriUtils.java
```diff
@@ -323,7 +323,7 @@ public static URI buildNewUrl(URI url, String relativeUrl)
                 throw new IllegalArgumentException(String
                         .format("Relative path '%s' for '%s' should start with forward slash ('/')", path, url));
             }
-            String uriAsString = createUriAsString(url.getScheme(), url.getAuthority(), encodeNonAsciiChars(path),
+            String uriAsString = createUriAsString(url.getScheme(), url.getRawAuthority(), encodeNonAsciiChars(path),
```
valfirst, line 326: please add unit tests

valfirst (changes requested):

valfirst (approved):

PR #4652, opened 2023-12-08
[plugin-web-app-to-rest-api] Add ability to use variable name as HTML…

uarlouski:
… source in from_html transformer

### DIFF
--- a/docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
+++ b/docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
@@ -250,7 +250,18 @@ Examples:
 |Description
 
 |`pageUrl`
-|The URL of the page to build the table upon.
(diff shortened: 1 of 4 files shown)


docs/modules/plugins/pages/plugin-web-app-to-rest-api.adoc
```diff
@@ -250,7 +250,18 @@ Examples:
 |Description
 
 |`pageUrl`
-|The URL of the page to build the table upon.
+a|The URL of the page to build the table upon.
+
+WARNING: The `pageUrl` parameter is deprecated, please use `variableName` instead.
```
valfirst, line 255: need to provide a version, when `pageUrl` is going to be removed

valfirst (changes requested):

valfirst (approved):

PR #4681, opened 2023-12-20
[CI] Use JDK 21 as runtime

valfirst:


valfirst:
Fix of PMD violations is required, because the number of total PMD violations for `vividus-engine` on JDK 21 is 21 (while max allowed value is 20).

PR #4682, opened 2023-12-20
[vividus-engine] Fix logging of deprecated steps without parameters inside composite steps

### LINKED ISSUE #4680: `NullPointerException` when run a composite step inside another composite step
`NullPointerException` when run a composite, for example, some deprecated step, inside another composite step. Only certain steps can cause such error. 
Note, that the execution of program doesn't stop, but in result the steps aren't executed and the report isn't generated.

**To Reproduce**
1. Create new composite step with deprecated step inside: 
```
Composite: When I run composite step
When I close the current window
```
2. Run new step (pay attention, to reproduce bug the deprecated step shouldn't be used outside of composite in current test run)

**Artifacts**
[log.txt](https://github.com/vividus-framework/vividus/files/13714073/log.txt)

**Execution environment:**
 - Vividus version: 0.6.4

draker94:
Closes #4680

### DIFF
--- a/vividus-engine/src/main/java/org/vividus/replacement/DeprecatedStepNotificationFactory.java
+++ b/vividus-engine/src/main/java/org/vividus/replacement/DeprecatedStepNotificationFactory.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Formatter;
 import java.util.List;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


vividus-engine/src/main/java/org/vividus/replacement/DeprecatedStepNotificationFactory.java
```diff
@@ -45,8 +46,8 @@ public DeprecatedStepNotificationFactory(RunTestContext runTestContext, Configur
     protected String createDeprecatedStepNotification(String versionToRemove, String replacementFormatPattern)
     {
         String runningDeprecatedStep = runTestContext.getRunningStory().getRunningSteps().getFirst();
-        Matcher stepMatcher = getStepMatcher(runningDeprecatedStep);
-        String[] stepParams = getParametersFromStep(stepMatcher);
+        Optional<Matcher> stepMatcher = getStepMatcher(runningDeprecatedStep);
+        String[] stepParams = stepMatcher.map(this::getParametersFromStep).orElseGet(() -> new String[0]);
```
valfirst, line 50: that's a fix of the reason, but not root cause, any plans to do root cause analysis?

draker94: > that's a fix of the reason, but not root cause, any plans to do root cause analysis?

The root of the problem is that steps within composite steps can be processed by separate logic (without using `stepMonitor` which we use to collect patterns of the current steps from tests) on the JBehave side.
https://github.com/jbehave/jbehave-core/blob/9510692f56f31551aa52e0eeddc03cc426deaea5/jbehave-core/src/main/java/org/jbehave/core/steps/StepCandidate.java#L242-L243

I'll think about how to make the "monitoring" of matched steps more unified on the JBehave side.

valfirst (approved):

PR #4697, opened 2023-12-28
[plugin-json] Step for validation against OpenAPI schema

abudevich:


vividus-plugin-json/src/main/java/org/vividus/json/steps/JsonSchemaValidationSteps.java
```diff
@@ -72,6 +75,24 @@ public void validateJsonAgainstSchema(String json, String schema)
         assertValidationMessages(validationMessages);
     }
 
+
+    @Then("JSON `$json` is valid against openapi schema `$openApiSchema`")
+    public void validateJsonAgainstOpenApiSchema(String json, String schema)
+    {
+        JsonNode componentNode = jsonUtils.readTree(schema).get(COMPONENTS);
+        JsonNode schemaNode = componentNode.get("schemas").elements().next();
```
valfirst, line 83: something wrong here, I believe it's not required to parse schema

vividus-plugin-json/src/main/java/org/vividus/json/steps/JsonSchemaValidationSteps.java
```diff
@@ -72,6 +75,24 @@ public void validateJsonAgainstSchema(String json, String schema)
...
+        ((ObjectNode) schemaNode).set(COMPONENTS, componentNode);
```
valfirst, line 84: `com.networknt.schema.JsonSchemaFactory#getSchema` ?

valfirst (changes requested):

vividus-plugin-json/src/main/java/org/vividus/json/steps/JsonSchemaValidationSteps.java
```diff
@@ -72,6 +75,24 @@ public void validateJsonAgainstSchema(String json, String schema)
...
+
+        SchemaValidatorsConfig config = new SchemaValidatorsConfig();
+        config.setTypeLoose(false);
+
+        JsonSchema jsonSchema = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7)
```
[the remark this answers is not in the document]
abudevich: As I understand, for OpenAPI 3.0 is valid to use v4/v6/v7 (https://json-schema.org/implementations#api-documentation).
But by https://swagger.io/docs/specification/data-models/keywords/#:~:text=OpenAPI%203.0%20uses%20an%20extended,and%20additional%20keywords%20are%20introduced. + https://json-schema.org/understanding-json-schema/reference/schema -> will be used v4.

vividus-plugin-json/src/main/java/org/vividus/json/steps/JsonSchemaValidationSteps.java
```diff
@@ -72,6 +75,24 @@ public void validateJsonAgainstSchema(String json, String schema)
...
+
+    @Then("JSON `$json` is valid against openapi schema `$openApiSchema`")
+    public void validateJsonAgainstOpenApiSchema(String json, String schema)
+    {
+        JsonNode componentNode = jsonUtils.readTree(schema).get(COMPONENTS);
+        JsonNode schemaNode = componentNode.get("schemas").elements().next();
+        ((ObjectNode) schemaNode).set(COMPONENTS, componentNode);
+
+        SchemaValidatorsConfig config = new SchemaValidatorsConfig();
+        config.setTypeLoose(false);
```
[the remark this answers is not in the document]
abudevich: during type validation if TYPE_LOOSE = true, will try to convert string to different types to match the type defined in schema

vividus-plugin-json/src/main/java/org/vividus/json/steps/JsonSchemaValidationSteps.java
valfirst: this comment is not resolved: the key problem here is that only one schema is retrieved from spec, what if the object to validate was created following the second or any other subsequent schema?

abudevich: Tested, for correct work of step need to have main dto as first in `schemas`.
As decision - we can pass main dto name as parameter in step and handle in inside step or modify schema with help path json step.

valfirst: it won't work if message entity is an array of DTOs

valfirst: what is the main goal of introducing this step? how will `$json` be provided?

abudevich: Usually it has nested structure.
Will not work for case with separate dto for error - because in common errorDto has another  structure in comparison with main response/request body.

As I understand, main goal - to validate potential response/request body agains schema which is generated by OpenApi spec.

PR #4741, opened 2024-01-19
[plugin-web-app-to-rest-api] Add support for jump links to named anchors in steps validating resources

### LINKED ISSUE #4735: Add support for anchor links `name` attribute in resource checker
The step ``Then all resources found by $htmlLocatorType `$htmlLocator` in $html are valid `` can check jump links (fetch the pointer from link and try to find the appropriate ID in document), but only if they using an ID as pointer.
But accordingly [the specification](https://www.w3.org/TR/html401/struct/links.html#h-12.2.1) jump links can use also and  `name` attribute as pointer.

Add ability to correctly check anchor links with the `name` attribute pointer.

Report with failed test during check the valid jump link:
[allure-jump-links.zip](https://github.com/vividus-framework/vividus/files/13976515/allure-jump-links.zip)

draker94:
Closes #4735

### DIFF
--- a/vividus-tests/src/main/resources/known-issues.json
+++ b/vividus-tests/src/main/resources/known-issues.json
@@ -93,7 +93,7 @@
   },
   "VVD-14": {
     "type": "Internal",
-    "assertionPattern": "Jump link by selector .* points to missing element with notFound id",
+    "assertionPattern": "Jump link by selector .* points to missing element with notFound id or name",
     "scenarioPattern": "Steps verification.*",
     "storyPattern": "ResourceCheckSteps",
     "stepPattern": "Then all resources found by.*"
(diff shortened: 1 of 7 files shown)


vividus-tests/src/main/resources/known-issues.json
```diff
@@ -93,7 +93,7 @@
   },
   "VVD-14": {
     "type": "Internal",
-    "assertionPattern": "Jump link by selector .* points to missing element with notFound id",
+    "assertionPattern": "(Jump link by selector .* points to missing element with notFound id or name|Element by selector .* doesn't contain href/src attributes)",
```
valfirst, line 96: why is the second part of the pattern needed?

draker94: > why is the second part of the pattern needed?

Now we have [named anchor](https://www.tagindex.net/html/link/a_name.html) on test page
https://vividus-test-site-a92k.onrender.com/links.html#ElementName

Hmmm, it would probably be better if we used locators that exclude such links

valfirst (approved):

PR #4910, opened 2024-03-18
[plugin-datetime] Add `fromEpochMilli`/`toEpochMilli` expressions

### LINKED ISSUE #4887: Add data/time fromEpochMillis capability
current functionality - [fromEpochSecond](https://docs.vividus.dev/vividus/latest/plugins/plugin-datetime.html#_fromepochsecond)

it could be nice to add the same as the existing functionality, but for Millis

**Describe alternatives you've considered**
I am using the following approach now #{round(#{eval(${variable}/1000)})} and then #fromEpochSecond

valfirst:
Closes #4887

### DIFF
--- a/docs/modules/plugins/pages/plugin-datetime.adoc
+++ b/docs/modules/plugins/pages/plugin-datetime.adoc
@@ -296,7 +296,8 @@ Calculates the difference between two dates with given formats and returns resul
 
 === `toEpochSecond`
 
-Converts the input date to the https://en.wikipedia.org/wiki/Unix_time[Unix epoch time] (the number of seconds from the epoch of 1970-01-01T00:00:00Z).
+Converts the input date to the number of seconds from the https://en.wikipedia.org/wiki/Unix_time[epoch]
+of 1970-01-01T00:00:00Z.
 
 [source, subs="+quotes"]
 ----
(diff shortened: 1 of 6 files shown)


docs/modules/plugins/pages/plugin-datetime.adoc
```diff
@@ -306,34 +307,69 @@ Converts the input date to the https://en.wikipedia.org/wiki/Unix_time[Unix epoc
...
+|734918400
+
 |`#{toEpochSecond(2020-12-11T18:43:05+05:30)}`
 |1607692385
 
+|`#{toEpochSecond(2020-12-11T18:43:05.987+05:30)}`
+|1607692385
+|===
+
+=== `toEpochMilli`
```
uarlouski, line 328: I believe it should be millis, not milli, toEpochSecond also seems to be wrong

uarlouski: but we don't expose java method names to users, so from user point of view millis/seconds look more relevant

valfirst: > but we don't expose java method names to users

true, but I was talking about naming, and my guess is Java naming is correct from English language standpoint

> so from user point of view millis/seconds look more relevant

I'm not native speaker, thus I can rely only on external sources (like, Java)

uarlouski (approved):

PR #4947, opened 2024-04-02
[plugin-applitools] Add ability to enable layout breakpoints in UFG

uarlouski:


docs/modules/plugins/pages/plugin-applitools.adoc
```diff
@@ -78,6 +78,10 @@ NOTE: The properties marked with *bold* are mandatory.
 |`false`
 |If `true`, page resources for rendering on the UFG will be fetched from outside of the browser.
 
+|`ui.visual.applitools.layout-breakpoints`
+|`false`
+|If `true`, the viewport width and height are changed to specified device's values prior to capturing the page source for rendering in UFG.
```
valfirst, line 83:
```suggestion
|If `true`, the viewport width and height are changed to the specified device values prior to capturing the page source for rendering in UFG.
```

valfirst (approved):

PR #4995, opened 2024-04-23
[tests] Debug failure in video tests

valfirst:


valfirst:
The issue with YouTube video is caused by changes on YouTube side: new rules against bots are introduced

PR #5029, opened 2024-05-04
[plugin-accessibility] Add ability to report a11y violations individu…

uarlouski:
…ally

### DIFF
--- a/vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
+++ b/vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019-2023 the original author or authors.
+ * Copyright 2019-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


vividus-plugin-accessibility/src/main/java/org/vividus/accessibility/AccessibilitySteps.java
```diff
@@ -126,14 +131,30 @@ public void performAccessibilityScan(ExamplesTable checkOptions)
...
+        {
+            violations.stream().map(v -> "[" + idGetter.apply(v) + "] " + messageGetter.apply(v))
+                    .forEach(softAssert::recordFailedAssertion);
+
+            if (!violations.isEmpty())
+            {
+                return;
+            }
+        }
+        softAssert.assertThat(String.format(format, args), (long) violations.size(), equalTo(0L));
```
valfirst, line 155:
```suggestion
        else
        {
            softAssert.assertThat(String.format(format, args), (long) violations.size(), equalTo(0L));
        }
```

valfirst, line 155:
```suggestion
        }
        else
        {
            softAssert.assertThat(String.format(format, args), (long) violations.size(), equalTo(0L));
        }
```

uarlouski: fixed, but it was intentional to perform this soft assert just to keep step non empty when there are no violations

uarlouski: merge pls :)

valfirst (approved):

PR #5033, opened 2024-05-10
[plugin-web-app] [wip] Update headless Chrome profile to use new headless mode

vkepin:


valfirst:
https://developer.chrome.com/blog/removing-headless-old-from-chrome

PR #5056, opened 2024-05-16
add support of mobitru doInjection parameter

dubovik-sergey:
Add support of doInjection parameter for installing native applications in the Mobitru.
This query parameter enables the injection of special code into an application to allow emulation of "touch id" action and QR code scan.

### DIFF
--- a/docs/modules/plugins/pages/plugin-mobitru.adoc
+++ b/docs/modules/plugins/pages/plugin-mobitru.adoc
@@ -125,7 +125,7 @@ a|`true` +
 a|`true` +
 `false`
 |`false`
-|inject special code into application to allow emulation of "touch id" action and QR code scan.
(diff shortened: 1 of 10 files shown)


docs/modules/plugins/pages/plugin-mobitru.adoc
```diff
@@ -121,6 +121,12 @@ a|`true` +
 |`true`
 |(iOS only) Resign the application (*.ipa) with Mobitru profile or not.
 
+|`mobitru.do-injection`
+a|`true` +
+`false`
+|`false`
+|inject special code into application to allow emulation of "touch id" action and QR code scan.
```
valfirst, line 128:
```suggestion
|Inject special code into application to allow emulation of "touch id" action and QR code scan.
```

valfirst (approved):

PR #5089, opened 2024-05-31
[azure-cosmos-db] Add ability to configure connection mode

abudevich:
Details: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/tune-connection-configurations-java-sdk-v4?tabs=api-sync#connection-configuration

### DIFF
--- a/docs/modules/plugins/pages/plugin-azure-cosmos-db.adoc
+++ b/docs/modules/plugins/pages/plugin-azure-cosmos-db.adoc
@@ -44,9 +44,11 @@ Where:
 
 . `account-key` - The key of the account configuration
 . `property-name` - The name of the container property. One of:
-.. `endpoint` - The endoint of the account.
+.. `endpoint` - The endpoint of the account.
 .. `key` - Either a master or readonly key used to perform authentication for Cosmos database service.
 If `key` is not set, <<_authentication,default authentication mechanism described below>> will be used.
+.. `connection-mode` - https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/sdk-connection-modes [The mechanism],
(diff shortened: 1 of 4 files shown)


docs/modules/plugins/pages/plugin-azure-cosmos-db.adoc
```diff
@@ -71,6 +73,7 @@ azure.cosmos-db.database.commecrial.account-key=vividus
 # Accounts configuration contains connection details
 azure.cosmos-db.account.vividus.endpoint=https://vividus-cosmos.documents.azure.com:443/
 azure.cosmos-db.account.vividus.key=some key
+azure.cosmos-db.account.vividus.connection-mode=direct,gateway
```
valfirst, line 76: I believe only one mode is allowed

vividus-plugin-azure-cosmos-db/src/main/java/org/vividus/azure/cosmos/CosmosDbService.java
```diff
@@ -66,6 +68,16 @@ public CosmosContainer load(CosmosDbContainer cosmosDbContainer)
...
+
+                        if (GATEWAY.equals(account.getConnectionMode()))
+                        {
+                            builder.gatewayMode();
+                        }
+                        else
+                        {
+                            builder.directMode();
+                        }
+
```
valfirst, line 80:
```suggestion

                        switch (account.getConnectionMode())
                        {
                            case GATEWAY -> builder.gatewayMode();
                            case DIRECT -> builder.directMode();
                        }

```

abudevich: Is it acceptable, if this will not cover case without specified `connection-mode` for CosmosDB account?

valfirst (approved):

PR #5115, opened 2024-06-10
Consider sorting of one element collection as passed

uarlouski:


valfirst (changes requested):
please update docs according to the changes

valfirst (approved):

PR #5124, opened 2024-06-12
[plugin-saucelabs] Set "VIVIDUS" value in "runner" meta-option of SauceConnect

valfirst:


uarlouski:
why do we need this?

valfirst:
> why do we need this?

by default SL SDK uses value "jenkins" in "runner" option, as the result users are confused (especially when running locally or using other CI systems)

uarlouski (approved):

PR #5134, opened 2024-06-17
use takeDeviceBySerial API in case if udid is provided in capabilities

SiarheiDubovik:


vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
```diff
@@ -63,6 +63,13 @@ public MobitruFacadeImpl(MobitruClient mobitruClient)
         this.mobitruClient = mobitruClient;
     }
 
+    @Override
+    public String takeDevice(String udid) throws MobitruOperationException
```
valfirst, line 67: please merge this logic into `org.vividus.mobitru.client.MobitruFacadeImpl#takeDevice(org.openqa.selenium.remote.DesiredCapabilities)`, use `DesiredCapabilities` object to determine which API to use to take device

valfirst (changes requested):

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
SiarheiDubovik: done

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
```diff
@@ -66,6 +67,20 @@ public MobitruFacadeImpl(MobitruClient mobitruClient)
     @Override
     public String takeDevice(DesiredCapabilities desiredCapabilities) throws MobitruOperationException
     {
+        Map<String, Object> capabilities = desiredCapabilities.asMap();
+        //use different API in case if udid is provided in capabilities
+        //it's required in some cases like if the device is already taken
+        if (!isSearchForDevice(desiredCapabilities)
+                &&
+                (capabilities.containsKey(APPIUM_UDID) || capabilities.containsKey(UDID)))
```
uarlouski, line 75: why not to move it else-if statement at line 89 so we don't need to potentially invoke isSearchForDevice twice?

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
```diff
@@ -66,6 +67,20 @@ public MobitruFacadeImpl(MobitruClient mobitruClient)
...
+        {
+            Object deviceIdObj = capabilities.getOrDefault(APPIUM_UDID, capabilities.get(UDID));
+            String udid = Optional.ofNullable(deviceIdObj).
+                    map(String::valueOf).
+                    orElseThrow();
```
uarlouski, line 80: afaik DesiredCapabilities drops all props with null values so its guaranteed to not be null

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
SiarheiDubovik: in order to avoid overcomplicated logic with multiple if else blocks, where it's easy to make a mistake

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
SiarheiDubovik: done

vividus-plugin-mobitru/src/main/java/org/vividus/mobitru/client/MobitruFacadeImpl.java
uarlouski: one if-else doesn't look overcomplicated :)

SiarheiDubovik: who may guarantee that there will be only one here forever?

uarlouski (approved):

Issue #4553, opened 2023-11-10
[Excel Plugin ]Add possibility to set or change the cell format when creating Excel files.


enerigo:
Please add possibility to set or change the **cell forma**t when creating **Excel** files. Right now, we're making an Excel file where everything is in General format. But now, we need to have some cells in Date format.

PR #5142, opened 2024-06-20
[plugin-excel] Add ability to preserve cells format

draker94:
Related to #4553

If we want to add cell typing it makes sense add ability to preserve cells format, otherwise, for example, dates, will be displayed as a number, and all numbers themselves will be displayed in fractional format no matter with which style we created the cell


vividus-plugin-excel/src/main/java/org/vividus/excel/transformer/ExcelTableTransformer.java
```diff
@@ -127,4 +128,9 @@ private List<String> replaceLineBreaks(List<String> list, String lineBreakReplac
                      .map(e -> e.replace("\n", lineBreakReplacement))
                      .toList();
     }
+
+    public void setPreserveCellFormatting(boolean preserveCellFormatting)
```
valfirst, line 132: please use constructor to set properties

vividus-plugin-excel/src/main/java/org/vividus/steps/ExcelDocumentValidationSteps.java
```diff
@@ -162,4 +163,9 @@ private static Predicate<Entry<CellValue, Optional<Pattern>>> filterMatched()
             return pattern.isPresent() || cellValue.value() != null;
         };
     }
+
+    public void setPreserveCellFormatting(boolean preserveCellFormatting)
```
valfirst, line 167: please use constructor to set properties

vividus-tests/src/main/resources/story/integration/TableTransformers.story
```diff
@@ -309,20 +309,6 @@ Then `${resultSequenceOfTransformers}` is equal to table:
...
-    @issueId 2451
-Then `<relativeUrl>` matches `.*links.*`
-Examples:
-{transformer=FROM_HEADLESS_CRAWLING, column=relativeUrl}
-
-
-Scenario: Verify FROM_HEADLESS_CRAWLING transformer passing URL through transformer parameter
-Then `<relativeUrl>` matches `.*links.*`
-Examples:
-{transformer=FROM_HEADLESS_CRAWLING, column=relativeUrl, mainPageUrl=$\{vividus-test-site-url\}}
```
valfirst, line 323: Why are these scenarios removed?

valfirst (changes requested):

vividus-tests/src/main/resources/story/integration/TableTransformers.story
draker94: Forgot to restore after local changes.

valfirst (approved):

PR #5166, opened 2024-06-27
[plugin-excel] Add handling for null cells

uarlouski:


vividus-tests/src/main/resources/story/integration/Excel.story
```diff
@@ -109,8 +109,9 @@ Meta:
 When I initialize scenario variable `expectedTable` with values:
 |StringValue|NumericValue|BooleanValue|FormulaValue|FormulaErrorValue|
 |City       |17          |FALSE       |289         |#VALUE!          |
+|Country    |19          |TRUE        |361         |null             |
```
valfirst, line 112:
```suggestion
|Country    |19          |TRUE        |361         |#{null}          |
```

valfirst (approved):

PR #5179, opened 2024-07-01
[plugin-web-app] Add steps to check cookie present using string comparison rules

draker94:


valfirst (commented):
Could you please update steps for playwright as well? (deprecation for playwright is not needed since steps have never been released)

valfirst (approved):

PR #5180, opened 2024-07-02
[plugin-web-app-playwright] Add step to upload file

avinBar:


vividus-plugin-web-app-playwright/build.gradle
```diff
@@ -22,6 +22,8 @@ dependencies {
     implementation(group: 'com.microsoft.playwright', name: 'playwright', version: '1.45.0')
     implementation platform(group: 'org.slf4j', name: 'slf4j-bom', version: '2.0.13')
     implementation(group: 'org.slf4j', name: 'slf4j-api')
+    implementation platform(group: 'org.springframework', name: 'spring-framework-bom', version: '6.1.10')
+    implementation(group: 'org.springframework', name: 'spring-context')
```
valfirst, line 26: is it really needed?

valfirst (changes requested):

vividus-plugin-web-app-playwright/build.gradle
avinBar: Otherwise, it complains that it cannot access the ResourceLoaderAware class file because org.springframework.context.ResourceLoaderAware was not found.

valfirst: Add an interface for `ResourceFileLoader`, e.g.

- `ResourceFileLoader` (interface) -> `SpringResourceFileLoader` (class)
- then use interface
- try to remove the dependency

valfirst (approved):

PR #5183, opened 2024-07-03
[vividus] Update regex for VaultStoredPropertiesProcessor

abudevich:


vividus/src/test/java/org/vividus/configuration/EncryptedPropertiesProcessorTests.java
```diff
@@ -50,7 +50,8 @@ class EncryptedPropertiesProcessorTests
     @CsvSource({
             ENCRYPTED_VALUE_PLACEHOLDER + ", " + DECRYPTED_VALUE,
             "username=open password=" + ENCRYPTED_VALUE_PLACEHOLDER + " secret-key=" + ENCRYPTED_VALUE_PLACEHOLDER
-                    + ", username=open password=" + DECRYPTED_VALUE + " secret-key=" + DECRYPTED_VALUE
+                    + ", username=open password=" + ENCRYPTED_VALUE_PLACEHOLDER
+                    + " secret-key=" + ENCRYPTED_VALUE_PLACEHOLDER
```
valfirst, line 54: that's a wrong change, properties are not decrypted anymore

abudevich: if it make on Vault processor level ?

valfirst (changes requested):

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
```diff
@@ -42,7 +42,7 @@ public class VaultStoredPropertiesProcessor extends AbstractPropertiesProcessor
 
     VaultStoredPropertiesProcessor(Properties properties)
     {
-        super("VAULT");
+        super("HASHI_CORP_VAULT|(?<!AZURE_KEY_)VAULT");
```
valfirst, line 45: user should see deprecation notification in logs, when old marker is used

vividus/src/main/java/org/vividus/configuration/AbstractPropertiesProcessor.java
```diff
@@ -23,13 +23,16 @@
...
+    private final String processorRegexMarker;
     private final Pattern propertyPattern;
 
-    AbstractPropertiesProcessor(String processorMarker)
+    AbstractPropertiesProcessor(String processorRegexMarker)
     {
-        this.processorMarker = processorMarker;
-        this.propertyPattern = Pattern.compile(processorMarker + "\\((.+?)\\)");
+        this.processorRegexMarker = processorRegexMarker;
+        this.propertyPattern = Pattern.compile(String.format(BRACKETS, processorRegexMarker) + VALUE_GROUP);
```
valfirst, line 35: why is it needed to wrap `processorRegexMarker` into group?

valfirst (changes requested):

vividus/src/main/java/org/vividus/configuration/AbstractPropertiesProcessor.java
abudevich: for correct work of regex for VAULT, didn't find better solution

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
```diff
@@ -66,6 +69,16 @@ protected String processValue(String propertyName, String partOfPropertyValueToP
                 ));
     }
 
+    @Override
+    public String processProperty(String propertyName, String propertyValue)
+    {
+        if (propertyValue.matches("(?<!AZURE_KEY_|HASHI_CORP_)VAULT\\((.+?)\\)"))
+        {
+            LOGGER.warn("Using 'VAULT(...)' is deprecated. Consider using 'HASHI_CORP_VAULT(...)' instead.");
```
valfirst, line 77: add unit tests to cover this line

valfirst, line 77:
```suggestion
            LOGGER.warn("`VAULT(...)` placeholder is deprecated and will be removed in VIVIDUS 0.7.0, please use `HASHI_CORP_VAULT(...)` placeholder instead.");
```

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
```diff
@@ -66,6 +69,16 @@ protected String processValue(String propertyName, String partOfPropertyValueToP
                 ));
     }
 
+    @Override
+    public String processProperty(String propertyName, String propertyValue)
+    {
+        if (propertyValue.matches("(?<!AZURE_KEY_|HASHI_CORP_)VAULT\\((.+?)\\)"))
```
valfirst, line 75: re-use `propertyPattern` here

vividus/src/systemTest/resources/properties/suite/valid/suite.properties
```diff
@@ -1,2 +1,2 @@
-property-stored-in-vault=VAULT(secret/vividus/test/top_secret)
-one-more-property-stored-in-vault=VAULT(secret/vividus/test/one_more_secret)
+property-stored-in-vault=HASHI_CORP_VAULT(secret/vividus/test/top_secret)
+one-more-property-stored-in-vault=HASHI_CORP_VAULT(secret/vividus/test/one_more_secret)
```
valfirst, line 2: we need to test both notations: new and deprecated

valfirst (changes requested):

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
abudevich: I can't reuse `propertyPattern` here, because it works for both `VAULT(...)` and `HASHI_CORP_ VAULT(...)`.
But log should be printed only for `VAULT(...)`

vividus/src/main/java/org/vividus/configuration/AbstractPropertiesProcessor.java
```diff
@@ -56,8 +56,8 @@ public String processProperty(String propertyName, String propertyValue)
         {
             String partOfValueToProcess = matcher.group(1);
             String processedPartOfValue = processValue(propertyName, partOfValueToProcess);
-            processedValue = processedValue.replace(processorMarker + "(" + partOfValueToProcess + ")",
-                    processedPartOfValue);
+            processedValue = processedValue.replace("(" + partOfValueToProcess + ")", processedPartOfValue)
+                    .replaceFirst(processorRegexMarker, "");
```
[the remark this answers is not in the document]
abudevich: updated, thanks

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
```diff
@@ -66,6 +69,17 @@ protected String processValue(String propertyName, String partOfPropertyValueToP
                 ));
     }
 
+    @Override
+    public String processProperty(String propertyName, String propertyValue)
+    {
+        if (propertyValue.matches("(?<!AZURE_KEY_|HASHI_CORP_)VAULT\\((.*?)\\)"))
```
valfirst, line 75: it should catch partial properties as well:
```
my-property=fooVAULT(secret-key)bar

File: vividus/src/main/java/org/vividus/configuration/AbstractPropertiesProcessor.java [hunk 9]
```diff
@@ -56,8 +54,7 @@ public String processProperty(String propertyName, String propertyValue)
         {
             String partOfValueToProcess = matcher.group(1);
             String processedPartOfValue = processValue(propertyName, partOfValueToProcess);
-            processedValue = processedValue.replace(processorMarker + "(" + partOfValueToProcess + ")",
-                    processedPartOfValue);
+            processedValue = processedValue.replaceFirst(propertyPattern.pattern(), processedPartOfValue);
```
valfirst, line 57: why first but not all?

vividus/src/test/java/org/vividus/configuration/VaultStoredPropertiesProcessorTests.java
```diff
@@ -156,4 +163,24 @@ void shouldUseNamespaceForEnterpriseVaultWhenConfigured() throws IOException
...
+
+    @ParameterizedTest
+    @ValueSource(strings = {"HASHI_CORP_VAULT()", "VAULT()"})
+    void shouldPrintLogAboutDeprecation(String expression) throws IOException
+    {
+        try (var processor = new VaultStoredPropertiesProcessor(new Properties()))
+        {
+            processor.processProperty("test-property-name", expression);
+        }
+        if ("VAULT()".equals(expression))
```
valfirst, line 175: unit tests should not contain business logic, in other words there should be no if-else etc. for this particular case it's required to split the test into 2 separate ones

valfirst (changes requested):

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
abudevich: and it works for such cases also:
Screenshot 2024-07-08 at 18 28 14

vividus/src/main/java/org/vividus/configuration/AbstractPropertiesProcessor.java
abudevich: because in `propertyPattern.pattern()` inside brackets is not specified particular value. 
as result - with `replaceAll()` will be replaced  not only needed expressions, which were processed.

vividus/src/main/java/org/vividus/configuration/VaultStoredPropertiesProcessor.java
valfirst: the screenshot is for "find", not for "matches"

vividus/src/test/java/org/vividus/configuration/VaultStoredPropertiesProcessorTests.java
```diff
@@ -156,4 +166,26 @@ void shouldUseNamespaceForEnterpriseVaultWhenConfigured() throws IOException
             ordered.verify(context).getBean(VaultTemplate.class);
         }
     }
+
+    @Test
+    void shouldPrintLogAboutDeprecation() throws IOException
+    {
+        try (var processor = new VaultStoredPropertiesProcessor(new Properties()))
+        {
+            processor.processProperty("old-property-name", "VAULT()");
```
valfirst, line 175: to reproduce comment, just try to change the test:

```suggestion
            processor.processProperty("old-property-name", "fooVAULT(secret-key)bar");
```

valfirst (changes requested):

valfirst (approved):

PR #5199, opened 2024-07-09
[plugin-web-app-playwright] Add scrolling steps

avinBar:


vividus-extension-web-app/src/main/java/org/vividus/ui/web/action/DirectionScroller.java
```diff
@@ -0,0 +1,33 @@
...
+
+    void scrollToBottom();
+
+    void scrollToLeft();
+
+    void scrollToRight();
+
+    static void unsupportedHorizontalScroll()
+    {
+        throw new UnsupportedOperationException("Horizontal scroll of the page not supported");
```
valfirst, line 31:
```suggestion
        throw new UnsupportedOperationException("Horizontal scroll of the page is not supported");
```

valfirst (changes requested):

PR #5223, opened 2024-07-17
[http-client] Decode credentials for Basic Authentication

draker94:
Currently HttpClient decode all characters except [unsafe](https://www.ietf.org/rfc/rfc1738.txt#:~:text=Other%20characters%20are%20unsafe%20because,be%20encoded%20within%20a%20URL.), but we don't need encodings of such characters when using Basic Auth


vividus-http-client/src/main/java/org/vividus/http/client/HttpClient.java
```diff
@@ -161,7 +162,8 @@ private static URI getRequestUri(ClassicHttpRequest request) throws IOException
     private static void configureBasicAuth(boolean usePreemptiveBasicAuthIfAvailable, String plainUserInfo,
             HttpHost host, HttpClientContext internalContext)
     {
-        UserInfo userInfo = UriUtils.parseUserInfo(plainUserInfo);
+        String plainUserInfoDecoded = URLDecoder.decode(plainUserInfo, StandardCharsets.UTF_8);
+        UserInfo userInfo = UriUtils.parseUserInfo(plainUserInfoDecoded);
```
valfirst, line 166: would you mind adding integration test for this case? you can use https://httpbin.org/#/Auth

valfirst (approved):

PR #5232, opened 2024-07-23
[plugin-web-app-playwright] Adjusted Playwright instantiation for par…

avinBar:
…allel execution

### DIFF
--- a/vividus-plugin-web-app-playwright/src/main/java/org/vividus/ui/web/playwright/BrowserContextProvider.java
+++ b/vividus-plugin-web-app-playwright/src/main/java/org/vividus/ui/web/playwright/BrowserContextProvider.java
@@ -31,20 +31,17 @@ import org.slf4j.LoggerFactory;
 import org.vividus.testcontext.TestContext;
 import org.vividus.ui.web.playwright.network.NetworkContext;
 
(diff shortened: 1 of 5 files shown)


vividus-plugin-web-app-playwright/src/main/java/org/vividus/ui/web/playwright/BrowserContextProvider.java
```diff
@@ -90,39 +96,14 @@ public void closeCurrentContext()
...
-                if (pc == null)
-                {
-                    Playwright playwright = Playwright.create();
-                    Browser browser = browserType.launchBrowser(playwright, launchOptions);
-                    pc = new PlaywrightContext(playwright, browser);
-                    playwrightContext = pc;
-                }
-            }
-        }
-        return pc;
```
valfirst, line 125: what is the root cause of these changes?

avinBar: The root cause is that we can't use the same playwright instance for all tests.
https://github.com/microsoft/playwright-java/issues/1184
https://playwright.dev/java/docs/multithreading

valfirst (approved):

PR #5289, opened 2024-08-09
[plugin-web-app] Deprecate `elementName` locator type

avinBar:


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/search/ElementNameSearch.java
```diff
@@ -35,6 +39,7 @@ public ElementNameSearch(LocatorType elementActionType)
     @Override
     public List<WebElement> search(SearchContext searchContext, SearchParameters parameters)
     {
+        LOGGER.warn("'elementName' locator type is deprecated  and will be removed in VIVIDUS 0.7.0");
```
valfirst, line 42: `name` locator uses this search as well, so this logging message should be removed

valfirst (approved):

PR #5416, opened 2024-10-03
[plugin-web-app-to-rest-api] Add ability to set checked attributes for resource checker

draker94:


vividus-plugin-web-app-to-rest-api/src/main/resources/properties/defaults/default.properties
```diff
@@ -19,5 +19,6 @@ transformer.from-headless-crawling.respect-no-follow=false
 transformer.from-headless-crawling.respect-no-index=false
 
 resource-checker.uri-to-ignore-regex=
+resource-checker.attributes-to-check=href,src
```
uarlouski, line 22: do we really need this property? isn't it going to be always either href or src?

draker94: > do we really need this property? isn't it going to be always either href or src?

1. Sites can have their own attributes to specify, for example, image source 
2. We may want to check only links (href) but not page resources (src)

valfirst: yes, nope

vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/ResourceCheckSteps.java
```diff
@@ -228,22 +232,25 @@ private static boolean isSchemaAllowed(URI uri)
...
 
-            String absUrl = element.absUrl(HREF_ATTR);
+            String absoluteUrl = element.absUrl(attributeWithValue);
             // For scripts e.g. href="javascript:alert('...');" the abs url will be empty
-            return absUrl.isEmpty() ? href : absUrl;
+            return absoluteUrl.isEmpty() ? hrefToCheck : absoluteUrl;
         }
-
-        return element.attr("src");
+        return attributeWithValue.isEmpty() ? StringUtils.EMPTY : element.attr(attributeWithValue);
```
valfirst, line 253:
```suggestion
        return attributeToCheck.map(element::attr).orElse(StringUtils.EMPTY);
```

vividus-plugin-web-app-to-rest-api/src/main/resources/properties/defaults/default.properties
uarlouski: but for a its always gonna be href

valfirst: some platforms don't follow specifications

valfirst (approved):

vividus-plugin-web-app-to-rest-api/src/main/resources/properties/defaults/default.properties
uarlouski: please provide markup example

PR #5423, opened 2024-10-04
[allure-adaptor] Add ability to configure report layout

valfirst:


docs/modules/configuration/pages/reporting.adoc
```diff
@@ -39,6 +39,122 @@ be used:
...
+a|`true` +
+`false`
+|`true`
+|Show or hide menu:Categories[] tab
+
+|`report.tabs.suites.enabled`
+a|`true` +
+`false`
+|`true`
+|Show or hide menu:Suites[] (aka menu:Batches[]) tab
```
draker94, line 54: image
Is this expected format of messages (I about '[]') ?
Shouldn't we add spaces after the colon?

valfirst: yes, it is, rendering is not correct without square brackets and no space is expected after colon: https://docs.asciidoctor.org/asciidoc/latest/macros/ui-macros/

PR #5547, opened 2024-11-26
[extension-selenium] Increase allowed published har size to 40 mb

uarlouski:


vividus-extension-selenium/src/main/java/org/vividus/ui/monitor/PublishingHarOnFailureMonitor.java
```diff
@@ -64,7 +66,16 @@ public PublishingHarOnFailureMonitor(boolean publishHarOnFailure, EventBus event
         this.testContext = testContext;
         this.mapperFactory = new MapperFactory()
         {
-            private final Supplier<ObjectMapper> objectMapper = Suppliers.memoize(ObjectMapper::new);
+            private static final int ALLOWED_HAR_SIZE = 40_000_000;
```
valfirst, line 69: 1) please add an information about exception which was hit before this fix
2) add the note to the documentation describing the current limit

valfirst (approved):

PR #5574, opened 2024-12-04
[for 0.7.0 release] Fix step to set browser cookies to HTTP context

vkepin:


vividus-plugin-web-app-to-rest-api/src/main/java/org/vividus/steps/integration/HttpRequestSteps.java
```diff
@@ -58,9 +54,11 @@ public HttpRequestSteps(CookieManager<Cookie> cookieManager, HttpTestContext htt
     public void executeRequestUsingBrowserCookies()
     {
         CookieStore cookieStore = cookieManager.getCookiesAsHttpCookieStore();
+        List<org.apache.hc.client5.http.cookie.Cookie> cookies = cookieStore.getCookies();
         attachmentPublisher.publishAttachment(ATTACHMENT_TEMPLATE_PATH,
-                Map.of(TEMPLATE_COOKIES_KEY, cookieStore.getCookies()), "Browser cookies");
-        httpTestContext.putCookieStore(cookieStore);
```
valfirst, line 63: I think it makes sense to remove the method `putCookieStore` from `HttpTestContext`

vividus-plugin-rest-api/src/main/java/org/vividus/http/HttpTestContext.java
```diff
@@ -54,11 +54,6 @@ public void addRequestHeaders(List<Header> requestHeaders)
         getData().requestHeaders.addAll(requestHeaders);
     }
 
-    public void putCookieStore(CookieStore cookieStore)
-    {
-        getData().cookieStore = cookieStore;
```
valfirst, line 59: if we can't it set, then it's always `null`, thus we can remove it, right?

valfirst (approved):

PR #5575, opened 2024-12-06
Add ability to disable scrollbar hiding

uarlouski:


vividus-plugin-web-app/src/main/java/org/vividus/selenium/screenshot/WebAshotFactory.java
```diff
@@ -95,10 +94,15 @@ private DebuggingViewportPastingDecorator decorateWithViewportPasting(ShootingSt
...
-            Optional<WebElement> scrollableElement)
+    private ShootingStrategy decorateWithScrollbarHiding(ShootingStrategy strategy, WebScreenshotParameters params)
     {
-        return new ScrollbarHidingDecorator(strategy, scrollableElement, scrollbarHandler);
+        if (params == null)
+        {
+            return new ScrollbarHidingDecorator(strategy, Optional.empty(), scrollbarHandler);
+        }
+        return params.isHideScrollbars()
+                ? new ScrollbarHidingDecorator(strategy, params.getScrollableElement(), scrollbarHandler)
```
valfirst, line 104: what if someone set: hideScrollbars = true, scrollableElement = "some_element"? it think there should be an error

uarlouski: scrollableElement is used to determine the height of a page under test, so these properties are not conflicting

PR #5597, opened 2024-12-13
[util] Escape path separators in batch variable keys

draker94:


vividus-util/src/main/java/org/vividus/util/property/PropertyMapper.java
```diff
@@ -42,6 +42,9 @@
 
 public class PropertyMapper implements IPropertyMapper
 {
+    private static final String VARIABLES_PROPERTY_PREFIX = "variables.";
```
valfirst, line 45: we can't hardcode it here:
1) `util` module should know nothing about `engine` module
2) this adds tight coupling to property naming

please check this: https://github.com/vividus-framework/vividus/commit/c9197cad2b2bdd7875fc7302638a9d2c451b8f7b

valfirst (changes requested):

PR #5696, opened 2025-02-07
Use crawler headers for redirects checking

uarlouski:


vividus-plugin-web-app-to-rest-api/src/main/resources/vividus-plugin/spring.xml
```diff
@@ -68,6 +67,25 @@
         <property name="filterRedirects" value="${transformer.from-sitemap.filter-redirects}" />
         <property name="mainPageUrl" value="${transformer.from-sitemap.main-page-url}" />
         <property name="mainPageUrlProperty" value="transformer.from-sitemap.main-page-url" />
+        <property name="httpRedirectsProvider">
+            <bean class="org.vividus.http.HttpRedirectsProvider">
+                <property name="httpClient">
```
valfirst, line 72: but we already have http client for `CrawlerCommonsSiteMapParser`, can't we reuse it?

valfirst (changes requested):

valfirst (approved):

PR #5928, opened 2025-06-06
Add abilit to save Lighthouse results path to a variable

uarlouski:


vividus-plugin-lighthouse/src/main/java/org/vividus/lighthouse/LighthouseSteps.java
```diff
@@ -360,8 +415,10 @@ private LighthouseResultV5 performLocalLighthouseScan(Path outputPath, String we
 
         try (InputStream stream = Files.newInputStream(outputPath))
         {
-            return pagespeedInsights.getObjectParser().parseAndClose(stream, StandardCharsets.UTF_8,
-                    LighthouseResultV5.class);
+            LighthouseResultV5 result = pagespeedInsights.getObjectParser().parseAndClose(stream,
+                    StandardCharsets.UTF_8, LighthouseResultV5.class);
+            result.put(FS_RESULT_PATH, outputPath.toString());
```
valfirst: it's not really correct to change the resulting object:
- it breaks consistency
- object in memory knows its location on the filesystem

it's better to change the returned result tyoe to entry: `Entry<Path, LighthouseResultV5>`

valfirst (changes requested):

vividus-plugin-lighthouse/src/main/java/org/vividus/lighthouse/LighthouseSteps.java
uarlouski: it would not be correct if we pass the object outside, however in our case the object is for internal use only, so to make to code clearer I believe such approach might be acceptable, using of Entry<Path, LighthouseResultV5> increases the complexity of the code due to common logic where Path is not required

valfirst: I think this line should be removed now

PR #5931, opened 2025-06-09
[plugin-lighthouse] Apply sorting before calculating percentiles

uarlouski:


uarlouski:
> no unit tests updates?

the sorting logic, that we just reuse, is already covered in unit test

PR #5970, opened 2025-06-27
Set focus state on field before typing text

uarlouski:


vividus-plugin-web-app/src/main/java/org/vividus/ui/web/action/FieldActions.java
```diff
@@ -228,6 +228,7 @@ private void sendKeysToElement(WebElement element, String text)
     {
         try
         {
+            javascriptActions.setFocusOnElement(element);
```
valfirst: doesn't `sendKeys` set the focus automatically?

valfirst (changes requested):

PR #6744, opened 2026-07-10
Add object pool steps

uarlouski:
## Summary by CodeRabbit

- **New Features**
  - Added object pool support with steps to initialize named pools, borrow pooled objects into scenario variables, and optionally return borrowed objects after each story completion.
  - Introduced `object-pool.return-objects-after-story-completion` (default: `false`) to control the return behavior.
- **Bug Fixes**
  - Improved JSON table transformation so object-typed column values are rendered as JSON in example tables.
- **Documentation**
  - Documented the new “Object pool” steps, including integration scenarios and nested-structure examples.
- **Tests**
  - Added coverage for pool lifecycle, variable storage, return-after-story behavior, and updated JSON expectations.

### DIFF
--- a/vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
+++ b/vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2019-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.vividus.steps;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
+
+import org.apache.commons.lang3.Validate;
+import org.jbehave.core.annotations.AfterStory;
+import org.jbehave.core.annotations.Given;
+import org.jbehave.core.annotations.When;
+import org.jbehave.core.model.ExamplesTable;
+import org.slf4j.Logger;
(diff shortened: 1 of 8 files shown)


vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @SuppressWarnings("PMD.AvoidSynchronizedAtMethodLevel")
+    private static final class ObjectPool
+    {
+        private final List<Map<String, String>> objects;
+
+        private ObjectPool(List<Map<String, String>> objects)
+        {
+            this.objects = new ArrayList<>(objects);
```
valfirst: it should be some concurrent list implementation

vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
```diff
@@ -0,0 +1,140 @@
...
+    {
+        if (returnObjectsAfterStory)
+        {
+            getTakenObjects().forEach((poolName, objects) -> objects.forEach(pools.get(poolName)::giveBack));
+        }
+    }
+
+    private Map<String, List<Map<String, String>>> getTakenObjects()
+    {
+        return testContext.get(TAKEN_OBJECTS_KEY, HashMap::new);
```
valfirst, line 120: what about marking the objects in pool as taken instead of creating a separate storage?

vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
```diff
@@ -0,0 +1,140 @@
...
+import org.vividus.testcontext.TestContext;
+import org.vividus.variable.VariableScope;
+
+public class ObjectPoolSteps
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(ObjectPoolSteps.class);
+
+    private static final Object TAKEN_OBJECTS_KEY = ObjectPoolSteps.class;
+
+    private final boolean returnObjectsAfterStory;
```
valfirst:
```suggestion
    private final boolean returnObjectsAfterStoryCompletion;
```

valfirst (changes requested):

vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
uarlouski: looks like overhead as we take random object, so every take is filtering of all objects by flag and then taking random from filtered, imho its simpler to have a separate storage, memory is cheaper than computation

vividus/src/main/java/org/vividus/steps/ObjectPoolSteps.java
uarlouski: synchronized is cheaper in this case than concurrent list implementations

valfirst (approved):

